PR #12, opened 2015-04-21
Update README, fix typos

btat:


brunoborges:
Thanks @btat for these!

PR #14, opened 2015-05-11
Update to Weblogic 12.1.3 update2 dev and OpenJDK 7u79

asziranyi:
weblogic and jdk update


brunoborges:
Thanks a lot for the latest update @asziranyi :-)

PR #22, opened 2015-07-06
Clarification when you use boot2docker

jmiddleton:
In the section where you mention how to access AdminServer Web Console, it would be better if you clarify that in case of using boot2docker, the ip address to access the console should be the output of `boot2docker ip` instead of the `{{NetworkSettings.IPAddress}}` of docker inspect.


brunoborges:
Would the authors of this PR be kind to provide a new PR that can be automatically merged?

PR #63, opened 2016-01-27
Revert "Name of the JRE download should be "server-jre-8u65-linux-x64.gz""

brunoborges:
Reverts oracle/docker-images#61


brunoborges:
The name of the file is with .tar.gz. Why do you suggest to rename it?

PR #72, opened 2016-02-12
Updated WebLogic dockerfiles to use JRE 8u73. 

mlvnd:
I also added .gitignore files to exclude the downloaded binaries from getting into the Git repo.


brunoborges:
Thanks for the submit, but we already had an update going on (pushed today: f3f62c6).

mlvnd:
Ok, thanks for the update.

PR #73, opened 2016-02-22
Fixed to set ADMIN_PASSWORD to an environmental valuable

charlier-shoe:
Fixed the Dockerfile to set ADMIN_PASSWORD to an environmental valuable, or else it fails to run "createServer.sh" and so on.
Because Managed Servers can't acquire ADMIN_PASSWORD and WLST call fails.


brunoborges:
Thank you for the submit. I've applied and tested the suggestion, and commited to 4820788

PR #84, opened 2016-04-05
New Docker sample for Tuxedo

DongHan2016:
hi Bruno,
I added the first sample of Tuxedo on Docker to GitHub. Would you please check?

Regards,
Dong


Djelibeybi:
Note that it's good practice to squash all your commits before submitting a pull request. That way, we don't have to store 32 individual commits in the master repo, just a single unified commit.

See https://davidwalsh.name/squash-commits-git for more info on how to squash your commits togher.

Djelibeybi:
Actually, this PR needs to be rejected and a squashed commit must be submitted, because the commit history contains binary files that we really don't want to store in the history.

brunoborges:
@DongHan2016, please see how we work with .download files as placeholders in the OracleWebLogic folder. Instruct users to download binaries from oracle.com and place them in the appropeiate folder.

Please submit a new PR with that approach instead.

PR #92, opened 2016-04-14
WLS Docker Multi-Host Servlet Proxy

mriccell:
Bruno
Could you please review my changes?

Thanks
Monica


OracleWebLogic/samples/1221-multihost/create-container.sh
```diff
@@ -7,7 +7,8 @@
 . ./setenv.sh
 
 uuid=$(uuidgen)
-name=$prefix-instance-$uuid
+#Define unique server name using suffix entered by user
+name=server$2
```
brunoborges: Hi @mriccell. Can you briefly describe why the change on name format for the container here?

OracleWebLogic/samples/1221-multihost/create-container.sh
```diff
@@ -23,6 +24,8 @@ fi
 eval "$(docker-machine env $swarm $machine)"
 
 docker run -d $DOCKER_CONTAINER_INSTANCE_OPTIONS \
+  -e MS_HOST=$name\
```
brunoborges: @mriccell, did you find any issue happening because these vars were not set? If yes, can you please provide some details? Just to get an understanding...

OracleWebLogic/samples/1221-multihost/bootstrap.sh
```diff
@@ -42,9 +42,19 @@ docker network create --driver overlay $network
...
+docker load -i _tmp_docker.img && rm _tmp_docker.img
+docker tag $adminimage 127.0.0.1:5000/$adminimage
+docker push 127.0.0.1:5000/$adminimage
+
+# Save existing defined image to a file to be loaded later into the registry created above
+eval "$(docker-machine env -u)"
 docker save $image > _tmp_docker.img 
 
-# Load, tag, and publish the image set in setenv.sh
+# Load, tag, and publish the managed server image app-deploy set in setenv.sh
```
brunoborges: @mriccell, I noticed that the defined image in setenv.sh has changed to `1221-servlet-proxy`, not `1221-app-deploy`. What is the real intention here? To use servlet-proxy or app-deploy image?

mriccell: I changed setenv.sh to have two images an $admin-image which represents the 1221-servlet-proxy (HttpProxyServlet) and $image which represents the 1221-appdeploy (sample app).
The 1221-servlet-proxy image is created by extending the 1221-appdeploy image. The HttpServletProxy (1221-servlet-proxy) application gets deployed to the admin server. The HttpServletProxy will route calls to applications deployed on the managed servers running in a WLS Cluster. The sample (1221-appdeploy) application gets deployed to servers in the WLS cluster DockerCluster.
When you make an HTTP call to an application using the admin server address (http://localhost:8001/sample) the HttpServletProxy will route the call to servers (server1 or server2) running in the WLS cluster.

OracleWebLogic/samples/1221-multihost/create-container.sh
mriccell: Yes I had to set the Managed Server host name so that I could access the servers from the HttpServletProxy application using a host name.  When we run the managed servers in a Docker Container we do not configure the listen address, thus we need to use a hostname to reach the servers.

OracleWebLogic/samples/1221-multihost/create-container.sh
mriccell: $name is used as the container host name and the server host name.  In the HttpClusterServlet web.xml, calls to the servers using the host name server1 or server2
  <servlet>
    <servlet-name>HttpClusterServlet</servlet-name>
    <servlet-class>weblogic.servlet.proxy.HttpClusterServlet</servlet-class>
    <init-param>
      <param-name>WebLogicCluster</param-name>
      <param-value>server1:7001|server2:7001</param-value>
    </init-param>
    <init-param>

brunoborges:
Based on existing sample with Apache WebLogic Plugin `1221-webtier`, this PR will be revised and submitted with the new approach.

PR #93, opened 2016-04-22
New Multi Host workshop using the Apache Plugin Web Tier 

mriccell:
Bruno

I have created a new workshop-multihost and have added a start-webtier script under 1221-multihost directory.  This new script will start an Apache Plugin Web Tier container in the weblogic-master machine.  After the WebLogic Domain is running in a multi host Docker environment we can invoke the sample application via the Apache Plugin Web Tier to any Managed Server in the WebLogic cluster.
Please review my changes at your earliest convinience.

Thanks
Monica


Djelibeybi:
I would strongly recommend squashing all your commits so that the history is not stored in the main repository. This is best practice when submitting a pull request so that only the result of your work is merged into the main repo, not each individual step.

Ref: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History (heading "Squashing Commits")

brunoborges:
@Djelibeybi luckily, GitHub provides the squash feature in a PR. :-)

Djelibeybi:
@brunoborges ah, yes. I haven't played with that yet. Though, I probably would've cleaned up the squashed commit message. :)

brunoborges:
My bad... :-)

PR #115, opened 2016-06-23
Enhancement for OHS readme(s).

hemastutibaruah:
Enhancement for OHS readme(s).


hemastutibaruah:
This contains the actual updates I Intended to put in README(s)

hemastutibaruah:
I have also added the sample env.list file require before docker run

brunoborges:
@hemastutibaruah so this is the right one?

PR #129, opened 2016-07-26
fix package name in OracleJDK image

rafael-pinho:
Instruction ADD fail during build because the search pattern for jre package is invalid. The jre package name ends with '.download'


brunoborges:
As described in the readme, you must download those files from oracle.com.

PR #139, opened 2016-08-02
Java alternatives

Djelibeybi:
This updates the base java-7 and java-8 builds to match what would occur if the Oracle JDK RPM is installed. This also leverages the `alternatives` tool to create system-wide in-path java/javac/jar binaries.


Djelibeybi:
I figured it was easier to comment on an actual pull request and I can just update my branch based on our discussions. I've modified things to match what happens when the JRE RPM is installed, but still using the server-jre tarball.

PR #145, opened 2016-08-12
Shrink image size

srt:
Shrink image size of Oracle Database XE image by a few hundred MB by skipping a layer with install files.


gvenzl:
With this change the second `COPY` instruction is redundant. The run file can also be copied in the above `COPY` instruction reducing the layers by one more: `COPY $INSTALL_FILE_1 $CONFIG_RSP $RUN_FILE $INSTALL_DIR/`

srt:
Thanks for the quick reply, I have updated the pull request.

brunoborges:
@gvenzl can you verify the suggested change?

gvenzl:
Changes verified, base image size reduces by about 500MB:

oracle/database     11.2.0.2-xe-pull    93cfc87ba81e        15 seconds ago      2.356 GB
oracle/database     11.2.0.2-xe         2e7c1807f054        6 minutes ago       2.839 GB

PR #149, opened 2016-08-23
Fix dockerfiles for weblogic

giggio:
There are no images `oracle/jdk` on the docker hub, only `oracle/openjdk`.


giggio:
The problem is that we get this error when building:
````
The OpenJDK JVM is not supported on this platform.
````

brunoborges:
We don't publish commercial products to Docker Hub. You have to build your own: https://github.com/oracle/docker-images/tree/master/OracleJDK

PR #172, opened 2016-10-06
Use the rpm to install java into the container

garthy:
This is an automated way to install java into a docker container. 
This also names the containers with a higher granularity oracle/jdk:8u101-b13 so you can specify a specific version in your FROM section.


brunoborges:
Even though technically it works, this would break the Oracle Java Binary Code License Agreement, because in practice the user building the image has no opportunity to read the license before accepting it, since the Dockerfile has the "accept" flag by default.

Thank you for the suggestion, but we cannot accept it.

PR #178, opened 2016-10-10
Simplify the java image build. Add more granular versions to the images

garthy:


brunoborges:
Thank you very much for your contribution. 

But it is already "complicated" (yeah...) for some folks on how to follow the README and drop the Server JRE tar.gz file in the same folder as the Dockerfiles. This would make things even more complicated (sad, I know...)

PR #180, opened 2016-10-11
Update README.md

eldermoraes:
Just to mention that the binaries don't need to be uncompressed in order to be used.

### DIFF
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -8,7 +8,7 @@ This project offers sample Dockerfiles for both Oracle Database 12c (12.1.0.2) E
 The `buildDockerImage.sh` script is just a utility shell script that performs MD5 checks and is an easy way for beginners to get started. Expert users are welcome to directly call `docker build` with their prefered set of parameters.
 
 ### Building Oracle Database Docker Install Images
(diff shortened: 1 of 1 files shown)


OracleDatabase/README.md
```diff
@@ -8,7 +8,7 @@ This project offers sample Dockerfiles for both Oracle Database 12c (12.1.0.2) E
 The `buildDockerImage.sh` script is just a utility shell script that performs MD5 checks and is an easy way for beginners to get started. Expert users are welcome to directly call `docker build` with their prefered set of parameters.
 
 ### Building Oracle Database Docker Install Images
-**IMPORTANT:** You will have to provide the installation binaries of Oracle Database and put them into the `dockerfiles/<version>` folder. You only need to provide the binaries for the edition you are going to install. You also have to make sure to have internet connectivity for yum.
+**IMPORTANT:** You will have to provide the installation binaries of Oracle Database and put them into the `dockerfiles/<version>` folder. You only need to provide the binaries for the edition you are going to install. You also have to make sure to have internet connectivity for yum. Note that you don't need to uncompress the binaries in order to use them, the script will handle it for you.
```
gvenzl: Actually, this should say that one must not uncompress the binaries otherwise unzip will throw error code 9 which leads the build being aborted. @eldermoraes, can you confirm that that was the case?

Issue #200, opened 2016-10-27
Oracle Database sample docker-compose file


montge:
Please provide sample docker compose file that includes persistent storage as sample.  This would be a good example to have handy as starting point for DBA's that are less familiar with docker, but work with teams using Docker Datacenter.

armand1m:
Here you go:

```yml
version: '2'
services:
  database:
    image: <internal-registry>/oracle/database:<tag> # e.g: registry.company.br/oracle/database:12.1.0.2-ee
    environment:
      - ORACLE_SID=xe
      - ORACLE_PDB=test
    volumes:
      - ./oracle/oradata:/opt/oracle/oradata # persistent oracle database data
      - ./data-bridge:/data-bridge # just to have someplace to put data into the running container if needed
    ports:
      - 1521:1521
      - 8080:8080
      - 5500:5500
```

ramonPires:
**A real example**
If you already have a database container running,that may be hard to recreate every time you run docker compose, and need to test something  like a dblink integration you can also specify in your docker-compose.yaml :
```
database:
#other stuff ....
    networks:
        - dbnet
    external_links: # https://docs.docker.com/compose/compose-file/#/externallinks
      - hard_to_create_oracle:oracle 
```
You also will need to add a network,because both containers need to be at same network:
`[user@machine]$ docker network create dbnet && docker network connect dbnet hard_to_create_oracle`
And add to network your docker-compose.yaml file:
```
version: '2'
services:
  database:
# service definition
networks: # https://docs.docker.com/compose/compose-file/#/networks
  dbnet:
      external: true
```
With docker composer you can create a network at your docker composer file,but you can't add to network a container which is not defined inside of your docker composer file,so that's why the network created outside of the docker composer file.

@armand1m I think your second solution will work.

armand1m:
I've actually tested the solution 2 today, and it doesn't really works.. Don't know if I'm doing something wrong, but I always need to create the `oradata` folder in the host system and execute  `sudo chown -R 1000:500 <path>` on it before running the container, as actually described in the Solution 1. If this is step is not done, the container is unable to start the oracle instance. It throws an error that calls SIGTERM, and then shuts down the container.

gvenzl:
A Docker Compose sample file is now provide within the samples folder: 12102-docker-compose

[...]

PR #214, opened 2016-11-21
Fix for #200 and others

gvenzl:
This PR includes:

- A docker compose example file for issue #200 
- Separation of database creation and startup into scripts
- Renamed example for starterset inside a container


brunoborges:
LGTM

PR #216, opened 2016-11-22
OracleDatabase: Decreased image size & support for VOLUME option.

lasjen:
Fork: [https://github.com/lasjen/docker-images](https://github.com/lasjen/docker-images)

This commit includes following changes:
1. Decreased image size

Using the COPY to transfer INSTALL_FILEs will generate large images. It does not help to remove the files afterwards in a RUN command. The files will still exist in the before-image.

Using docker to run a http file server, the files can be downloaded locally with a WGET command in the same RUN command as removing the files.

2. Making the VOLUME an option
Makes it much easier to create an image with a created database on docker internal filesystem (local). This is achieved by creating the image first without VOLUME. If VOLUME option set, a VOLUME is added to the image.

3. Refactoring code
The script is refactored by generating several functions.

4. Updated OracleDatabase/README.md
Updated README.md regarding changes in 1 and 2 (above).


OracleDatabase/dockerfiles/12.1.0.2/Dockerfile.ee
```diff
@@ -75,6 +79,8 @@ RUN sed -i -e "s|###ORACLE_EDITION###|EE|g" $INSTALL_DIR/$INSTALL_RSP &&
 USER oracle
 
 RUN cd $INSTALL_DIR       && \
```
clemSeveillac: If we want to squash the big intermediate layers, which should be removed indeed, shouldn't we either:
- convert once and only once e.g. via the build.sh the 2 install zip files to one big .tar.gz file, that we can add and untar at the right location via a simple ADD command here?
- or be patient for docker 1.13, which will add the ability (albeit in experimental mode) to squash all layers into one? :) (in the meantime, you can docker run <yourBigImage> true then docker export | docker import it, and add metadata again, but I agree it's tedious)

That way, we wouldn't have to impose a new restriction (sudo) on that build system...

brunoborges:
I believe we shouldn't be optimizing the image size through complex building steps that depend on external elements (i.e. http server running separately to serve install files).

Although I like the idea and I myself did a POC with the exact approach, I found that it would remove the ability for users to build images without using a helper shell script (buildDockerImage.sh). It would enforce a dependency. Plus, as @clemSeveillac stated, Docker itself will soon provide a capability to easily squash layers.

@gvenzl, thoughts?

lasjen:
Hmm ... I don't know if it is right to considering docker to be an external and complex building step, when we do actually work with docker :-). But ... I do understand what you mean, Bruno. 

I thought most people was actual using the buildDockerimage.sh. Advanced users could easily edit the Dockerfile. But of course - I do see the point regarding ruining manual builds. 

Thanks for the info. I did not know Docker was planning to include an option to merge images (in version 1.13). That is great news.

I have made a branch "size_decr_w_httpserv" that I´ll use for my builds, and reverted my forked master. Have some other ideas I want to share.

gvenzl:
I agree with the above comments. There are a lot of ways to further optimize the image and image size but with each optimization another piece of complexity is added that limited the general applicability of the image. Same goes with the optional inclusion of the VOLUME instruction.

Users with deep knowledge of Docker can create their own build files to work around issues that they have, as you have done @lasjen. But there are other uses with not so deep knowledge about Docker and all they look for is a simple to build and run Docker image.

Issue #226, opened 2016-12-01
Sometime openssl rand -base64 8 generated password does not conform to the Oracle recommended standards


marcelo-ochoa:
runOracle.sh script uses **_openssl rand -base64 8_** command to generate a random password, but some time it doesn't included a number inside, for example **_LhCsOGTOaTM=_** warning generated is:

> [WARNING] [DBT-06208] The 'SYS' password entered does not conform to the Oracle recommended standards.
>    CAUSE: 
> a. Oracle recommends that the password entered should be at least 8 characters in length, contain at least 1 uppercase character, 1 lower case character and 1 digit [0-9].
> b.The password entered is a keyword that Oracle does not recommend to be used as password
>    ACTION: Specify a strong password. If required refer Oracle documentation for guidelines.

is not a big problem but can confuse non experimented users :)

brunoborges:
@gvenzl perhaps appending a 0 at the end just to make sure it has a number would be sufficient...

gvenzl:
Yes, that is probably a good approach. Note, the auto generated password can always be reset with `docker exec <container name> ./setPassword.sh <your password>`.

gvenzl:
This issue is now fixed, a "1" will be appended to the password automatically.

PR #231, opened 2016-12-11
Simplification of Dockerfiles, fix for #226, Docker version output during build

gvenzl:


ManFromEast:
could you help me to stop that receive email from pull request？

PR #243, opened 2016-12-28
Add function to set database initialization parameters

chiroito:
You will be able to set the initialization parameters of the database by this change.

### DIFF
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -48,6 +48,7 @@ To run your Oracle Database Docker image use the **docker run** command as follo
 	-p <host port>:1521 -p <host port>:5500 \
 	-e ORACLE_SID=<your SID> \
 	-e ORACLE_PDB=<your PDB name> \
+	[-e INIT_PARAMS=<adding init params>] \
 	-v [<host mount point>:]/opt/oracle/oradata \
 	oracle/database:12.1.0.2-ee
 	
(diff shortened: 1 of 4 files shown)


gvenzl:
What benefit does this deliver that cannot be achieved by setting the parameter directly once the database container is up?

clemSeveillac:
@gvenzl I see some params like `db_unique_name` that cannot be changed after DB creation, and that are quite important if you want to set up two containers with Data Guard for instance.

gvenzl:
That parameter can be controlled via `-e ORACLE_SID=<YOUR DB>` when starting the container.

clemSeveillac:
The SID can, the `db_unique_name` cannot; in my case Primary and Standby share the same SID while they have different `db_unique_name`s... But there are many other changes that must be done to init scripts anyway to make primary-ready and standby-ready images, i.e. this PR would only address a small part :)

gvenzl:
The trouble I have with this is that if users can just randomly set init parameters the DB creation might fail because of that. Of course that would be the users responsibility but the reality is that this will lead to a lot of new GitHub issues from not so advanced users and will leave a bad taste in their mouths. I rather have a simple, ready to go Oracle DB in Docker that addresses the needs of 90% of the users out there -> "Give me an Oracle DB in Docker", rather than trying to put in every edge case that will just make the build more complicated.
Advanced users can always build their own Dockerfile from the official one, one of the benefits that the DB is created on container startup and not fix in the image, and do with it whatever they want.
We even simplified that by providing shell scripts to group instructions so that some of this is reusable.
If there is a parameter that we are missing and that makes a lot of sense, I'm happy to expose it via `-e` directly but just giving users the ability to set any init.ora parameter doesn't sound right to me. I just can't see the need for that, hence why I'm asking why you want this. :)

brunoborges:
+1 to keeping Dockerfiles simple. Advanced users should just tweak their own files.

Gr33nbl00d:
i think this is really missing! Its quite hard to fiddle out how to create a database with a specific init.ora. So bad this was simply closed

gvenzl:
@Gr33nbl00d, what stops you from extending the image and build a database specific to your needs or use the `/opt/oracle/scripts/setup` functionality to set the parameters you need in the spfile?

PR #247, opened 2017-01-03
Update index.html

cloudrivers:


gvenzl:
I don't see any added benefit for this pull request. This was merged originally by accident and then reverted back.

Issue #246, opened 2016-12-30
Cannot start OracleDB image


ggrimbert:
Hi,

I have create a docker image for Oracle DB 12.1.0.2 EE.

When I run the following command 
```
docker run \
-p 1521:1521 -p 5500:5500 \
-e ORACLE_SID=sequelize \
-e ORACLE_PDB=test_sequelize \
-v /home/kimgit/oracle:/opt/oracle/oradata \
oracle/database:12.1.0.2-ee
```

I got the following logs : 
```
ORACLE AUTO GENERATED PASSWORD FOR SYS, SYSTEM AND PDBAMIN: LuZvF9abF3U=1

LSNRCTL for Linux: Version 12.1.0.2.0 - Production on 30-DEC-2016 09:51:28

Copyright (c) 1991, 2014, Oracle.  All rights reserved.

Starting /opt/oracle/product/12.1.0.2/dbhome_1/bin/tnslsnr: please wait...

TNSLSNR for Linux: Version 12.1.0.2.0 - Production
System parameter file is /opt/oracle/product/12.1.0.2/dbhome_1/network/admin/listener.ora
Log messages written to /opt/oracle/diag/tnslsnr/e786bd61093f/listener/alert/log.xml
Listening on: (DESCRIPTION=(ADDRESS=(PROTOCOL=ipc)(KEY=EXTPROC1)))
Listening on: (DESCRIPTION=(ADDRESS=(PROTOCOL=tcp)(HOST=0.0.0.0)(PORT=1521)))

Connecting to (DESCRIPTION=(ADDRESS=(PROTOCOL=IPC)(KEY=EXTPROC1)))
STATUS of the LISTENER
------------------------
Alias                     LISTENER
Version                   TNSLSNR for Linux: Version 12.1.0.2.0 - Production
Start Date                30-DEC-2016 09:51:28
Uptime                    0 days 0 hr. 0 min. 0 sec
Trace Level               off
Security                  ON: Local OS Authentication
SNMP                      OFF
Listener Parameter File   /opt/oracle/product/12.1.0.2/dbhome_1/network/admin/listener.ora
Listener Log File         /opt/oracle/diag/tnslsnr/e786bd61093f/listener/alert/log.xml
Listening Endpoints Summary...
  (DESCRIPTION=(ADDRESS=(PROTOCOL=ipc)(KEY=EXTPROC1)))
[...]

PR #218, opened 2016-11-29
fix wrong log path

twonky4:
If dbca failed the dbca-log should be displayed, but the specified path is wrong. I triggered this error by using an SID with invalid character.

PR #250, opened 2017-01-05
Include valid SID check (fix issue #246)

gvenzl:
Checks:
- That SID is not longer than 12 bytes
- That SID is alphanumeric only


clemSeveillac:
Thanks! Doesn't this make this little PR obsolete? https://github.com/oracle/docker-images/pull/218 (it was about `dbca` logs alternate location when SID is invalid)

gvenzl:
Sort of, but it isn't entirely clear when the log file is located in a sub directory named after the SID and when not. Therefore I hesitate to remove this PR as the command is only executed anyway if the previous one failed, due to the `||` bash operator.

clemSeveillac:
Makes sense!

PR #262, opened 2017-01-30
Database: Provide a env variable to set the character set during db creation

michael-simons:
I'd kindly propose the following change:

* Introduce `ORACLE_CHARACTERSET` defaulted to AL32UTF8
* Change the dbca template to have a place holder for CHARACTERSET instead of the fix value above
* Use an env variable to determine the characterset for the database

Thus, the character set can be passed to the container on the first run to create a database with a different characterset for Standard Edition 2 and Enterprise Edition.

I have updated the README accordingly and hope this is of use to you too.

### DIFF
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -39,6 +39,8 @@ Before you build the image make sure that you have provided the installation bin
 
 You may extend the image with your own Dockerfile and create the users and tablespaces that you may need.
 
+The character set for the database is set during creating of the database. 11g Express Edition supports only UTF-8. You can set the character set for the Standard Edition 2 and Enterprise Edition during the first run of your container and may keep separate folders containing different tablespaces with different character sets.
+
 ### Running Oracle Database in a Docker container
 
 #### Running Oracle Database Enterprise and Standard Edition in a Docker container
@@ -48,6 +50,7 @@ To run your Oracle Database Docker image use the **docker run** command as follo
 	-p <host port>:1521 -p <host port>:5500 \
 	-e ORACLE_SID=<your SID> \
 	-e ORACLE_PDB=<your PDB name> \
+	-e ORACLE_CHARACTERSET=<your character set> \
 	-v [<host mount point>:]/opt/oracle/oradata \
 	oracle/database:12.1.0.2-ee
 	
@@ -57,6 +60,8 @@ To run your Oracle Database Docker image use the **docker run** command as follo
 	                  Two ports are exposed: 1521 (Oracle Listener), 5500 (OEM Express)
 	   -e ORACLE_SID: The Oracle Database SID that should be used (default: ORCLCDB)
 	   -e ORACLE_PDB: The Oracle Database PDB name that should be used (default: ORCLPDB1)
+	   -e ORACLE_CHARACTERSET:
+	                  The character set to use when creating the database (default: AL32UTF8)
 	   -v             The data volume to use for the database.
 	                  Has to be owned by the Unix user "oracle" or set appropriately.
 	                  If omitted the database will not be persisted over container recreation.
--- a/OracleDatabase/dockerfiles/12.1.0.2/createDB.sh
+++ b/OracleDatabase/dockerfiles/12.1.0.2/createDB.sh
@@ -34,6 +34,7 @@ cp $ORACLE_BASE/$CONFIG_RSP $ORACLE_BASE/dbca.rsp
 sed -i -e "s|###ORACLE_SID###|$ORACLE_SID|g" $ORACLE_BASE/dbca.rsp
 sed -i -e "s|###ORACLE_PDB###|$ORACLE_PDB|g" $ORACLE_BASE/dbca.rsp
 sed -i -e "s|###ORACLE_PWD###|$ORACLE_PWD|g" $ORACLE_BASE/dbca.rsp
+sed -i -e "s|###ORACLE_CHARACTERSET###|$ORACLE_CHARACTERSET|g" $ORACLE_BASE/dbca.rsp
 
 # Create network related config files (sqlnet.ora, tnsnames.ora, listener.ora)
 mkdir -p $ORACLE_HOME/network/admin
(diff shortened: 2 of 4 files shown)


brunoborges:
This falls under the category _nice to have, but it is yet another variable that can be easily tweaked by users customizing their own images_

michael-simons:
No. That is essentially part of the user data. We had this discussion about size, and having another ~12Gb image just for providing the charset? i really think that one is important. It cannot be changed after the db is created. 

Then what about an option to generally provide the DBCA.rsp by the user? Thus i wouldn't have to have an image for each var and you wouldn't have to deal with those vars.

gvenzl:
What's the use case for having a different character set other than AL32UTF8, given that AL32UTF8 is a superset of all Oracle supported character sets? AL23UTF8 has become the recommended default for exactly that reason, to avoid character set conversions and conversations. Why do you need to have a database in WE8MS1252 if your application will just run fine with AL32UTF8 instead. Or is that not the case?

michael-simons:
No, we have actually several old applications that depend on WE8MS1252 with table names having umlauts. Yes that is bad and all but I cannot change that. I cannot use their dumps with table spaces generated with Al32UTF8.

brunoborges:
@michael-simons it seems you have this need for "your" scenario. We want to make these Docker images as extensible and customizable as possible for anyone, and yet usable at first without complicated settings.

I believe you can simply write your own Dockerfile and customize what you need, and then build your own base Docker image for DB, and push to a private registry server for others to reuse.

michael-simons:
Yeah sure. Complicated settings. The parameter is defaulted. Please take at least a look. Nothing changed for using it out of the box. 

Everyone has their need for his own scenario. 

Your believe here is clearly wrong.

michael-simons:
On a 2nd @brunoborges your argument applies much better to the SID parameters. Why bother to allow changing them through complicated settings at all? The database is isolated in a container and I dare to say the SID has no or at least less impact than the charset for an application running against a db inside that container.  

Back to the PR: it's documented, tested, doesn't change the previous behavior at all. What's so bad about sharing "my" usecase of that container? Isn't that what open source is about, too?

gvenzl:
But even umlauts in table names, column names, etc. are possible with AL32UTF8, still don't understand why that character set would impose an issue on that. 

You are the first to bring this up so I would like to understand the issue you are facing, first.

michael-simons:
Assuming that there is no other way we can import our dump, the proposed change fixes the problem that the character set of the database cannot be changed easily by:
* Having a container that can create databases / tablespaces with any supported charset
* Starting the container as before with an empty volume it leads to creation of standard database
* Starting the container with an empty volumn and providing the charset, the database is created with that (unchangable) charset
* further starts of a container pointing to the same folder don't need any parameter (as they don't need params for the sid), the charset is part of the folder that contains the tablespace
The image itself isn't really changed by this!

We could extend the image but than there would be another 12Gb for changing ALU32UTF8 to WE8MS1252… 

Yet another, better solution would be: Allow to pass dbca.rsp.tmpl during the start of the image. That way a user could modify *all* vars (which are partially not changeable after the creation of a tablespace). If no dbca.rsp.tmpl is given, fine, use the default one.

silke-boehler:
Hi together,
I'm a colleague of @michael-simons. We were having those problems with characterset conversion, because we have object names with german umlauts that are 30 charcters long. When exporting those from MSWIN1252 and importing to ALU32UTF8, the umlauts will expand to 2 byte and the objects names will not fit any more into 30 bytes.

michael-simons:
Thanks @silke-boehler 

Here is the relevant part from the database documentation:

> In the database data dictionary, schema object names cannot exceed 30 bytes in length. You must rename schema objects if their names exceed 30 bytes in the new database character set. For example, one Thai character in the Thai national character set requires 1 byte. In AL32UTF8, it requires 3 bytes. If you have defined a table whose name is 11 Thai characters, then the table name must be shortened to 10 or fewer Thai characters when you change the database character set to AL32UTF8.

See: "Additional Problems Caused by Data Truncation" https://docs.oracle.com/database/121/NLSPG/ch11charsetmig.htm#NLSPG461 Object names always use byte characteristics, not char.

So we think that having the database always in ALU32UTF8 is sufficient.

michael-simons:
Here's one for the lols: "freileitung_erdkabel_übergänge" (also, you cannot use it with Workspace manager, but that's another topic)

gvenzl:
Thanks @michael-simons and @silke-boehler, this makes sense to me now. Indeed, if you are already using up to 30 bytes or are close to, then the 2 byte conversion for the German umlauts does of course cause a problem. Just wanted to get this confirmed as we are recommending to use AL32UTF8 going forward so that hopefully one day these conversion conversations will be a problem of the past. I know.. famous last words  :)

Btw, if you haven't known yet, we have increased the limit to 128 bytes for all objects in 12.2 (table names, column names, procedure names, etc). That should also help your case going forward to adopt AL32UTF8 and still have enough space for the 2 bytes for the umlauts. Of course similar problem will occur if somebody decides to use all the 128 bytes to come up with a very, very, very explanatory name for the table or column name. But let's just hope that common sense kicks in before that.

As to being able to provide a custom dbca.rsp.tmpl, this is something that we would not want to enable. The primary goal for these Docker images is to make it quick and easy for people to use an Oracle Database. Many users already have shown that they are not necessarily experts when it comes to administrate a database, and that is perfectly fine. By providing too many knobs that can be turned we will likely just end up with many more issues stating that the db isn't working while the problem is really an improper config. For people who do have the technical skills they are welcome to build their own image from this base image with providing their own dbca.rsp.tmpl file if wanted or needed.
By keeping this images as simple as possible we guarantee the applicability to the widest possible user base.

Your fix works fine. Note, when an invalid character set is specified, such as the above stated `WE8MS1252` which really should be `WE8MSWIN1252`, the DB defaults back to `US7ASCII`. I'm fine with that, there are too many possible character sets which would make it too cumbersome to test for validity before starting the DB creation. If the users knows what he is doing, he ends up with the right character set. If he doesn't, chances are that he hasn't change the setting in the first place and ends up with `AL32UTF8`, the recommended default. And if that is not the case the DB will still be created with a default character set of `US7ASCII`.

Issue #270, opened 2017-02-08
DBCA loses contact during first run


neerolyte:
I'm not sure why this has started happening, I upgraded Docker (to 1.13.0 on OS X) and now when I run an Oracle container it fails pretty early during DBCA:

```
$ docker run -it --name oracle-db oracle/database:12.1.0.2-ee
ORACLE AUTO GENERATED PASSWORD FOR SYS, SYSTEM AND PDBAMIN: kjFwWpzHVNA=1

LSNRCTL for Linux: Version 12.1.0.2.0 - Production on 08-FEB-2017 20:00:54

Copyright (c) 1991, 2014, Oracle.  All rights reserved.

Starting /opt/oracle/product/12.1.0.2/dbhome_1/bin/tnslsnr: please wait...

TNSLSNR for Linux: Version 12.1.0.2.0 - Production
System parameter file is /opt/oracle/product/12.1.0.2/dbhome_1/network/admin/listener.ora
Log messages written to /opt/oracle/diag/tnslsnr/6d4cfaec1cf6/listener/alert/log.xml
Listening on: (DESCRIPTION=(ADDRESS=(PROTOCOL=ipc)(KEY=EXTPROC1)))
Listening on: (DESCRIPTION=(ADDRESS=(PROTOCOL=tcp)(HOST=0.0.0.0)(PORT=1521)))

Connecting to (DESCRIPTION=(ADDRESS=(PROTOCOL=IPC)(KEY=EXTPROC1)))
STATUS of the LISTENER
------------------------
Alias                     LISTENER
Version                   TNSLSNR for Linux: Version 12.1.0.2.0 - Production
Start Date                08-FEB-2017 20:00:55
Uptime                    0 days 0 hr. 0 min. 0 sec
Trace Level               off
Security                  ON: Local OS Authentication
SNMP                      OFF
Listener Parameter File   /opt/oracle/product/12.1.0.2/dbhome_1/network/admin/listener.ora
Listener Log File         /opt/oracle/diag/tnslsnr/6d4cfaec1cf6/listener/alert/log.xml
Listening Endpoints Summary...
  (DESCRIPTION=(ADDRESS=(PROTOCOL=ipc)(KEY=EXTPROC1)))
  (DESCRIPTION=(ADDRESS=(PROTOCOL=tcp)(HOST=0.0.0.0)(PORT=1521)))
The listener supports no services
The command completed successfully
Copying database files
1% complete
3% complete
DBCA Operation failed.
[...]

PR #275, opened 2017-02-11
ensure errors in installPerl.sh are caught

neerolyte:
Makes #270 a lot easier to spot.


gvenzl:
Thanks Dave,

It was me thinking that you could see the review of the line. But seems like this is for repo owners only. However, your explanation makes sense to me now.

Thanks a lot for your contribution, Oracle highly appreciates it and so do I!

PR #279, opened 2017-02-16
Auto generate Node Manager password in HTTPS domain image

mriccell:
Instead of setting the Node Manager password as a environment variable, auto generate the Node Manager password when the container is run.


mriccell:
@MattAtTTU I have changed the Dockerfile and container-scripts to auto generate the NM password

PR #296, opened 2017-02-24
Remove yum install util-linux

mriccell:
Remove yum install util-linux and remove su to make image as small as possible.
Move common commands to the head of Dockerfiles to reutilize layers across different WLS version images.
Standardize the use of port numbers, admin port 7001, managed server port 7002

### DIFF
--- a/OracleWebLogic/dockerfiles/12.1.3/Dockerfile.developer
+++ b/OracleWebLogic/dockerfiles/12.1.3/Dockerfile.developer
@@ -40,14 +40,27 @@ FROM oracle/serverjre:8
 # ----------
 MAINTAINER Bruno Borges <bruno.borges@oracle.com>
 
-# Environment variables required for this build (do NOT change)
-# -------------------------------------------------------------
-ENV FMW_PKG=wls1213_dev_update3.zip \
-    MW_HOME=/u01/oracle \
-    ORACLE_HOME=/u01/oracle \
+# Environment variables required for this build
+# ---------------------------------------------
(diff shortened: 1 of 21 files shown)


Djelibeybi:
I suggest a few minor tweaks:

- Update `12.2.1/Dockerfile.infrastructure` as well
- Move the "Setup filesystem and oracle user" section in all the `Dockerfiles` to after the `MAINTAINER` directive so that the layer created will be shared between all WLS images.

The second one means that you need to add this block at the top of the `Dockerfile` across all versions:

```
RUN chmod a+xr /u01 && \
    useradd -b /u01 -m -s /bin/bash oracle && \
    echo oracle:oracle | chpasswd && \
    chown oracle:oracle -R /u01 
```

It also means you'll have to add a new `RUN` directive later to `chmod +xr $SCRIPT_FILE` after you copy in the `$SCRIPT_FILE`.

Djelibeybi:
LGTM. Thoughts from @brunoborges or @gvenzl?

mriccell:
@Djelibeybi @brunoborges @gvenzl I am in the process of testing these images

Djelibeybi:
I've asked @mriccell to squash all these commits so that our review is a bit simpler.

mriccell:
@Djelibeybi  @brunoborges @gvenzl I have squash the commits and they are ready for review

gvenzl:
LGTM

Djelibeybi:
Merged. Thanks @mriccell :)

PR #303, opened 2017-03-04
Oracle SOA 12c Docker Image

jeqo:
Scripts to create Docker image for SOA 12.2.1.2 based on JRE server 8 and sample to create a OSB compact domain.


OracleSOA/dockerfiles/12.2.1.2/Dockerfile.quickstart
```diff
@@ -0,0 +1,40 @@
+# Pull base image
+# ---------------
+FROM oracle/serverjre:8
```
Djelibeybi: Can you review the new structure for the WebLogic `Dockerfile` and try and re-order this `Dockerfile` to match? It looks like there is a lot of similarity between the two and layer re-use can be increased if we can match directives. 

It will also allow future SOA releases to re-use layers.

OracleSOA/samples/12212-osb-compact-domain/container-scripts/create-osb-domain.py
```diff
@@ -0,0 +1,69 @@
+domain_name  = os.environ.get("DOMAIN_NAME", "base_domain")
```
Djelibeybi: Does this Python script need a shebang `#!/usr/bin/python` line?

OracleSOA/dockerfiles/12.2.1.2/Dockerfile.quickstart
jeqo: Yes, it's based in WebLogic version. 
With reorder do you mean to have an initial domain once we start the container? (i.e. add the 'createAndStartEmptyDomain.sh') How will the base domain looks like? SOA Suite has at least 3 main componets: SOA, BPM and OSB, the empty domain will have only SOA? Why not keep an 'intermediate' layer with the product installed and have additional images based on that one, one for each main component?

I think we can probably reuse the infrastructure layer to create a SOA "generic" image.

OracleSOA/samples/12212-osb-compact-domain/container-scripts/create-osb-domain.py
jeqo: Not sure, but I'll add the license headers as in WebLogic samples.

OracleSOA/dockerfiles/12.2.1.2/Dockerfile.quickstart
Djelibeybi: No, I mean take a look at the new WebLogic Dockerfiles: @mriccell did a great job to restructure them so that there is more layer re-use between versions. I think the same techniques could be used here.

OracleSOA/samples/12212-osb-compact-domain/container-scripts/create-osb-domain.py
klauern: I believe the shebang line indicates what executable to run, so if it had it, wouldn't that break the way it works?  I assume that is a WLST script meant to be run in Jython, so specifying `/usr/bin/python` would probably fail, correct?

jeqo: It is executed by WLST, even adding the shebang works as expected

OracleSOA/dockerfiles/12.2.1.2/Dockerfile.quickstart
jeqo: Updated.

jeqo:
@Djelibeybi anything else that could be improved to merge this PR?

Djelibeybi:
It's waiting on @brunoborges and @mriccell to review as they are the WLS domain experts. Also needs confirmation of the OCA.

Djelibeybi:
I'm going to close the PR as the SOA team have indicated that they're in the process of building an official SOA Docker image.

jeqo:
Cool :+1:

PR #304, opened 2017-03-06
Fix bug in createAndStartEmptyDomain.sh

chrisguoado:
On starting the 12.2.1.2 container, below error is logged:
`cat: date: No such file or directory`

Then the calculated admin password will always be `d41d8cd9`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/*.jar
 **/*.zip
 **/*.tar.gz
+.idea/
--- a/OracleWebLogic/dockerfiles/12.2.1.2/container-scripts/createAndStartEmptyDomain.sh
+++ b/OracleWebLogic/dockerfiles/12.2.1.2/container-scripts/createAndStartEmptyDomain.sh
@@ -38,7 +38,7 @@ fi
 # Create Domain only if 1st execution
 if [ $ADD_DOMAIN -eq 0 ]; then
 # Auto generate Oracle WebLogic Server admin password
-ADMIN_PASSWORD=$(cat date| md5sum | fold -w 8 | head -n 1) 
+ADMIN_PASSWORD=$(date| md5sum | fold -w 8 | head -n 1)
 
 echo ""
 echo "    Oracle WebLogic Server Auto Generated Empty Domain:"


chrisguoado:
BTW, I didn't see there's any Oracle DB running inside the container, while there's below lines at the top of the script:

```bash
########### SIGTERM handler ############
function _term() {
   echo "Stopping container."
   echo "SIGTERM received, shutting down database!"
   sqlplus / as sysdba <<EOF
   shutdown immediate;
EOF
   lsnrctl stop
}

########### SIGKILL handler ############
function _kill() {
   echo "SIGKILL received, shutting down database!"
   sqlplus / as sysdba <<EOF
   shutdown abort;
EOF
   lsnrctl stop
}
```

Djelibeybi:
Assigning to @mriccell to check/test.

mriccell:
@chrisguoado that is to turn off the javadb which is started by default

gvenzl:
@mriccell that must be a typo then. `sqlplus / as sysdba` certainly only connects to an Oracle DB. Perhaps the instruction were forgotten to be changed within the handlers?

.gitignore
```diff
@@ -2,3 +2,4 @@
 **/*.jar
 **/*.zip
 **/*.tar.gz
+.idea/
```
Djelibeybi: What's this for?

mriccell:
@chrisguoado 
I am not sure how these DB commands made it into the createAndStartEmptyDomain.sh, they are unnecessary for WLS.  I will remove them in a pull request.

.gitignore
mriccell: @Djelibeybi the .gitignore is so that when we do a git commit it does not include any of the installers that might be in the directory

Djelibeybi: Yeah, I know what `.gitignore` does. :) I meant, what creates the `.idea/` directory and why is it being added to `.gitignore`? This seems like it could be something from a git IDE that's not actually required.

chrisguoado: It's the IntelliJ IDEA project configuration, something like `.settings` for a Eclipse project. I think it's common to clone a GitHub repo into a new IDEA project, and we definitely don't want to commit the files under `.idea/`.

PR #313, opened 2017-03-08
Provide PowerShell script for Windows platform

fjhorrillo:
Bash script translate to Windows PowerShell

### DIFF
--- a/OracleDatabase/dockerfiles/buildDockerImage.ps1
+++ b/OracleDatabase/dockerfiles/buildDockerImage.ps1
@@ -1,4 +1,4 @@
-#PowerShell
+#!/bin/powershell
 # 
 # Since: March, 2017
 # Author: fjhorrillo@gmail.com
--- a/OracleDatabase/dockerfiles/buildDockerImage.ps1
+++ b/OracleDatabase/dockerfiles/buildDockerImage.ps1
@@ -79,7 +79,7 @@ function checksumPackages() {
     exit $?
   }
 }
-s
+
 ##############
 #### MAIN ####
 ##############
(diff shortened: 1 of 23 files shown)


Djelibeybi:
This will require an OCA, so I'm assigning @brunoborges to handle that, and @gvenzl can review the script itself.

fjhorrillo:
I have added one more script for WebLogic and fixed a bug about reading Checksum file

fjhorrillo:
I have added scripts for OracleJava

gvenzl:
Removing the label `database` as there are more scripts provided now. What I'm unclear, Mac user speaking: Will be a powershell script be enough, just to make building the images easier, or does Windows also require a change of the Dockerfiles themselves. I.e. right now this is running under the Linux 7 slim image. Do we need to use a Windows image and install the Windows based Database binaries?

fjhorrillo:
Hi gvenzl, my scripts allow user to build the same as in Linux (using Linux containers) with the same syntax but from Windows (I tested it on Windows 10).
Other option is what you mean, using nanoserver (Windows Server 2016 nano image) instead of Linux but for that it is necessary to change the Dockerfiles and the result would no longer be the same images, it will be the equivalent in the Windows platform.

gvenzl:
Thanks a lot for the clarification, @fjhorrillo

brunoborges:
@Djelibeybi @gvenzl I can't review/test/maintain these scripts. Can some of you? If we can't find someone in our group to do that, I'm afraid we won't be able to take these in

Djelibeybi:
@brunoborges I don't have access to any Windows systems. I cannot review/test/maintain these scripts.

Djelibeybi:
I'm closing this PR due to lack of activity and lack of internal access to Windows for QA purposes.

PR #320, opened 2017-03-15
Added support for auto-deploying Java applications on boot of WebLogic

craigbarrau:
Prior to this PR, there was no way to easily build custom WebLogic images based off the official Docker images. This PR allows a user to do the following:

Given a Dockerfile of
```
FROM container-registry.oracle.com/middleware/weblogic:12.2.1.1
COPY sample.war .
```
And a `sample.war` file in the build context folder `.`
When the user runs `docker build -t sample .`
And then user runs `docker run -ti -p 7001:7001 sample`
Then they will have a running WebLogic domain with the sample app pre-deployed

On boot, it works by scanning `/u01/oracle` directory for `*.war` and `*.ear` files and deploying each one via WLST offline after the domain is created but before it is booted. Obviously, if no applications exist it will just boot a vanilla WebLogic domain as per usual.

As the 12.2.1.1 and 12.2.1.2 Docker images are the only ones that create a working WebLogic domain, they are the only ones where this approach will work. If 12.1.3 image is updated in future to create a WebLogic domain on boot, this same solution would work.

### DIFF
--- a/OracleWebLogic/dockerfiles/12.2.1.1/Dockerfile.developer
+++ b/OracleWebLogic/dockerfiles/12.2.1.1/Dockerfile.developer
@@ -51,7 +51,7 @@ RUN mkdir -p /u01 && \
    
 # Copy scripts
 #-------------
-COPY container-scripts/createAndStartEmptyDomain.sh container-scripts/create-wls-domain.py /u01/oracle/
+COPY container-scripts/createAndStartEmptyDomain.sh container-scripts/create-wls-domain.py container-scripts/deploy-app.py /u01/oracle/
 
 ENV SCRIPT_FILE=/u01/oracle/createAndStartEmptyDomain.sh \
     DOMAIN_NAME="${DOMAIN_NAME:-base_domain}" \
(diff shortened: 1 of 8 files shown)


mriccell:
@craigbarrau we do not want to deploy application in images under dockerfile, these are only install images which create an empty domain!  Under samples we have the 1221-appdeploy that extends the image created in the step above and deploys the application.  Another way is copying your application to the $DOMAIN_HOME autodeploy directory and when the server is started the application will be deployed.  I probably will create a sample for this.

craigbarrau:
Nice one Monica! I did not know about `autodeploy`. 

I tried this and it worked perfectly...
```
FROM container-registry.oracle.com/middleware/weblogic:12.2.1.1
COPY sample.war $DOMAIN_HOME/autodeploy/sample.war
```

Thanks heaps Monica - love your work!

Cheers,
Craig

mriccell:
Hi Craig,

Thank you so much for the feed back!  Do you mind creating a pull 
request for this?  The dockerfile and scripts should be under the 
samples sub directory and that your FROM is not from the registry but 
from a build image from GitHub oracle/weblogic:12.2.1.1-generic for example.

Thanks
Monica

PR #330, opened 2017-03-23
Use "exec" where possible to avoid excess shell processes

tianon:
As written, this ends up keeping a `/bin/sh` process resident as PID 1, which can cause issues with signals being sent to the service, but mostly is unnecessary. :+1: :heart:

In case you aren't familiar with `exec`, here's the relevant docs from Bash: (https://www.gnu.org/software/bash/manual/bashref.html#index-exec)

> If command is supplied, it replaces the shell without creating a new process. ... If command cannot be executed for some reason, a non-interactive shell exits, unless the execfail shell option is enabled. In that case, it returns failure. ...

Happy to edit, rebase, adjust, etc as desired!

### DIFF
--- a/NoSQL/sample-cluster/inner-scripts/Dockerfile
+++ b/NoSQL/sample-cluster/inner-scripts/Dockerfile
@@ -4,4 +4,4 @@ RUN mkdir sample-cluster
 
 COPY *.sh ./sample-cluster/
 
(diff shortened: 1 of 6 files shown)


Djelibeybi:
Thanks @tianon, merged.

PR #333, opened 2017-03-27
Oracle TSAM Plus on Docker

chrisguoado:
Dockerfiles and scripts to create Docker image for Oracle TSAM Plus 12.2.2.1.


OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+    TMPFILES=/tmp/files
+
+# Setup filesystem and oracle user
+# Adjust file permissions, go to /u01 as user 'oracle' to proceed with WLS installation
+# ------------------------------------------------------------
+RUN mkdir -p /u01/oracle && \
+    chmod a+xr /u01 && \
+    useradd -u 1000 -b /u01 -M -s /bin/bash oracle && \
+    chown oracle:oracle -R /u01 && \
+    yum install -y libaio vim tar net-tools openssh-server rsync sudo && \
```
Djelibeybi: Why are you installing OpenSSH, rsync and sudo? These should not be necessary inside a Docker container.

OracleTuxedo/OracleTSAM/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+This project offers Dockerfile for building:
+ * Oracle TSAM Plus 12c Release 2 (12.2.2) + Rolling Patch 003
+
+## Dependencies
+This project depends on the [Oracle Server JRE 8 Docker Image](../..//OracleJava). So, before you proceed to build this image, make sure the image `oracle/serverjre:8` has been built locally or is accessible in a remote Docker registry.
+
+**IMPORTANT:** You will have to provide the required installation binaries and put them into the `dockerfiles/<version>` folder.
+
+For this image following installation media / binaries are required:
```
Djelibeybi: You don't provide a link to the exact binaries required. This should be added.

OracleTuxedo/OracleTSAM/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+* Oracle TSAM Plus 12.2.2 GA Installer
+* Oracle TSAM Plus 12.2.2 Rolling Patch 003
+* Oracle Database Instant Client
```
Djelibeybi: You should require and install the RPM version of the Instant Client.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+#
+# REQUIRED FILES TO BUILD THIS IMAGE
+# ----------------------------------
+# (1) Oracle Tuxedo System and Applications Monitor (TSAM) Plus 12cR2 (12.2.2) GA Installer: tsam122200_64_Linux_x86.zip
+#     Download from http://www.oracle.com/technetwork/middleware/tuxedo/downloads/index.html
+#
+# (2) Oracle TSAM Plus 12.2.2 Rolling Patch 003: p25448043_12220_Linux-x86-64.zip
+#     Download from https://updates.oracle.com/Orion/Services/download/p25448043_12220_Linux-x86-64.zip?aru=20968945&patch_file=p25448043_12220_Linux-x86-64.zip
+#
+# (3) Oracle Database Instant Client: instantclient-basiclite-linux.x64-12.2.0.1.0.zip & instantclient-sqlplus-linux.x64-12.2.0.1.0.zip
```
Djelibeybi: You should not use the zip files. Rather, switch to using the Instant Client RPMs on Oracle Linux

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+# ------------------------------------------------------------
+USER oracle
+RUN cd /u01 && \
+      mkdir oraInventory && \
+      cd /u01 && \
+      jar xf $FMW_PKG && \
+      cd /u01/Disk1/install && \
+      chmod -R +x * && \
+      sh ./runInstaller.sh -silent -waitforcompletion -responseFile /u01/tsam_install.rsp; \
+      if [ $? = 255 ];then \
```
Djelibeybi: Can you explain this more? It seems like you try an install and if it fails you do exactly the same install again?

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+      /u01/oracle/oraHome/OPatch/opatch apply 25448043.zip && \
+      cd && rm -rf tmp /u01/$RP_PKG
+
+WORKDIR /u01/oracle
+
+USER root
+COPY files /tmp/files/
+RUN mkdir /etc/envs && \
+      cp $TMPFILES/bashrc /etc/envs && \
+      bash -c "echo . /etc/envs/bashrc >> /etc/bashrc" && \
```
Djelibeybi: Why are you creating `bashrc` files? Docker containers rarely run `bash`.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/files/sbin/init.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+source ~/sbin/functions.sh
+source /etc/envs/bashrc
+
+nohup sudo /sbin/sshd -D >/dev/null 2>&1 &
```
Djelibeybi: Why are you running `sshd` inside a container? How are you controlling this pid?

Djelibeybi:
Ping @brunoborges and @gvenzl for additional review.

brunoborges:
@chrisguoado sent you a note over email

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: OK, no problem, I'll switch to RPMs.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: `rsync` was to facilitate the development, and is not used in this image, I can remove it.
`sudo` is for starting the `sshd`.
While running on a public cloud (like OCCS), `ssh` with public key auth is usually the only option to connect into the container.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: Yes, they are exactly the same.
We found that when running the OUI inside a container which is hosted by a VirtualBox VM, the first  run usually fails with a `Error in CreateOUIProcess(): 26` error. A second try always works.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: The initialize script of this image depends heavily on `bash` and some functions in this `bashrc` file. I think `bash` is the only and default shell in `oraclelinux:7-slim`. Please explain more on "Docker containers rarely run bash", I don't quite follow you.

OracleTuxedo/OracleTSAM/README.md
chrisguoado: There is a description "Please refer to the **REQUIRED FILES TO BUILD THIS IMAGE** section in the [Dockerfile](dockerfiles/12.2.2/Dockerfile) for more detail." right after those lines. All the required information are in the `Dockerfile`:

```
# REQUIRED FILES TO BUILD THIS IMAGE
# ----------------------------------
# (1) Oracle Tuxedo System and Applications Monitor (TSAM) Plus 12cR2 (12.2.2) GA Installer: tsam122200_64_Linux_x86.zip
#     Download from http://www.oracle.com/technetwork/middleware/tuxedo/downloads/index.html
#
# (2) Oracle TSAM Plus 12.2.2 Rolling Patch 003: p25448043_12220_Linux-x86-64.zip
#     Download from https://updates.oracle.com/Orion/Services/download/p25448043_12220_Linux-x86-64.zip?aru=20968945&patch_file=p25448043_12220_Linux-x86-64.zip
#
# (3) Oracle Database Instant Client: instantclient-basiclite-linux.x64-12.2.0.1.0.zip & instantclient-sqlplus-linux.x64-12.2.0.1.0.zip
#     Download from http://www.oracle.com/technetwork/topics/linuxx86-64soft-092277.html
```

Does this work?

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/files/sbin/init.sh
chrisguoado: I did't think the pid (of the `sshd` process) should be paid special care, should I?

brunoborges: We follow best practices to all images. That said, please remove the SSH deamon

OracleTuxedo/OracleTSAM/README.md
brunoborges: Create ".download" files as you see them in the OracleWeblogic folder for example. Please follow the same pattern.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
brunoborges: Oracle WebLogic and Oracle Database Docker images will trigger OUI in silent mode. It never fails. Please double check your scripts. If there is a bug, ensure the bug is raised to the OUI team.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
brunoborges: Cloud providers may give SSH access into the VMs running the images, or they implement bridges for `docker exec`. Please remove sshd (and consequently, sudo).

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
Djelibeybi: If you rely on `bash`, you should call it directly, i.e. via `/bin/bash` not via `/bin/sh`. Also, you shouldn't rely on any shell defaults. if your initialize script requires specific functions, then you should define those and call them specifically from the initialize script. There is no guarantee that the shell environment is going to read the defaults.If you rely on `bash`, you should call it directly, i.e. via `/bin/bash` not via `/bin/sh`. Also, you shouldn't rely on any shell defaults. if your initialize script requires specific functions, then you should define those and call them specifically from the initialize script. There is no guarantee that the shell environment is going to read the defaults.

Further, your `.bashrc` is 1036 lines long, which is massive. Do you really need all those functions and aliases? I would review that and strip it back down to exactly what's needed.

OracleTuxedo/OracleTSAM/README.md
Djelibeybi: The `README.md` should contain all the links and you should include `.download` files as @brunoborges mentioned that also contains the links. Our top support issue is users not properly downloading the binaries before building images, so we want as much documentation for this as possible.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/files/sbin/init.sh
Djelibeybi: There is no `init` inside a Docker container, so you need to be careful about child processes you spawn without reaping. By `nohup`'ing `'sudo` you make it more difficult to introspect the process inside the container or reap it properly.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: OK, will remove sshd and and sudo.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: Since it happens only in a OEL container running inside a VirutalBox VM, I'll remove the second try.

Djelibeybi: (It's Oracle Linux or OL. Never OEL).

Also, if you can reproduce it regularly on OL inside a VirtualBox VM, please log it as a bug so it can be investigated by OL/VirtualBox/OUI developers to determine _why_ it's failing.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: I'll re-define the `ENTRYPOINT` in Dockerfile to call `/bin/bash` explicitly.
It's not the `.bashrc` that large, but the `bashrc` file copied into `/etc/envs` folder. It is working as a common library, and there's indeed many function unused by this image. If it matters, I'll remove the redundant ones.

OracleTuxedo/OracleTSAM/README.md
chrisguoado: OK, I see.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/files/sbin/init.sh
chrisguoado: OK, `sshd` and `sudo` will be removed.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: OK.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
```diff
@@ -100,10 +100,8 @@ RUN mkdir /etc/envs && \
       bash -c "echo . /etc/envs/bashrc >> /etc/bashrc" && \
       bash -c "echo . /etc/envs/bashrc >> /u01/oracle/.bashrc" && \
       cd /u01/oracle && \
-      jar xf /u01/$ORACLT_PKG1 && \
-      jar xf /u01/$ORACLT_PKG2 && \
-      chmod -R +x instantclient_* && \
-      mv /u01/oracle/instantclient* /u01/oracle/instantclient && \
+      rpm -i --nodeps /u01/$ORACLT_PKG1 /u01/$ORACLT_PKG2 && \
```
Djelibeybi: Please do not use the `rpm` utility to install packages. This shuold be replaced by `yum localinstall` so that package dependencies are resolved and installed as well.

Installing with `--nodeps` is very dangerous and should never be done.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
```diff
@@ -100,10 +100,8 @@ RUN mkdir /etc/envs && \
...
+      mkdir /u01/oracle/instantclient && \
```
Djelibeybi: No need to still create this directory. The previous command would install the Instant Client for you.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: Updated to use `yum localinstall`.
Just wanted to use basiclite version of instant client to make the downloads / image size as small as possible. But neither `rpm -i basiclite.rpm sqlplus.rpm` nor `yum installlocal basiclite.rpm sqlplus.rpm` let me install. It keep complaining that dependency is not satisfied, which looks like a bug. So I had to use `--nodeps` option as I know `libaio` is the only dependency and I had it.

OracleTuxedo/OracleTSAM/dockerfiles/12.2.2/Dockerfile
chrisguoado: This `~/instantclient` directory is now just a custom directory containing two scripts, see this:
```bash
mv $TMPFILES/sqlrun/* /u01/oracle/instantclient
```

Djelibeybi:
@chrisguoado are you still going to reduce the size of `.bashrc`? Or is this ready for another review?

chrisguoado:
@Djelibeybi I'm working on it, and will be ready later today. As well as adding the download info into README and creating .download files.

Djelibeybi:
@chrisguoado great, thanks. Drop a comment here when you're ready for us to review, so we don't start while you're still working. :)

chrisguoado:
@Djelibeybi It's done. Till now all the existing comments should have been addressed.

OracleTuxedo/OracleTSAM/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Dependencies
+This project depends on the [Oracle Server JRE 8 Docker Image](../..//OracleJava). So, before you proceed to build this image, make sure the image `oracle/serverjre:8` has been built locally or is accessible in a remote Docker registry.
+
+**IMPORTANT:** You will have to provide the required installation binaries and put them into the `dockerfiles/<version>` folder.
+
+For this image following installation media / binaries are required:
+
+* [Oracle TSAM Plus 12.2.2 GA Installer](http://www.oracle.com/technetwork/middleware/tuxedo/downloads/index.html) - `tsam122200_64_Linux_x86.zip`
+* [Oracle TSAM Plus 12.2.2 Rolling Patch 004](https://updates.oracle.com/Orion/Services/download/p25530287_12220_Linux-x86-64.zip?aru=21140450&patch_file=p25530287_12220_Linux-x86-64.zip) - `p25530287_12220_Linux-x86-64.zip`
```
Djelibeybi: Is this available to customers only? If so, we can't make it a requirement for installation. We need to restrict this repo to binaries that are publicly available.

chrisguoado: You mean the rolling patch? I wasn't aware of that. Need to confirm it...

chrisguoado: @Djelibeybi Yes, you are right. Just confirmed the rolling patches are only for customers. So, I'll need to do some adjustment to the Dockerfile and README to remove the RP as a requirement.

chrisguoado: However, the reason why we were recommending RP003 is, in that patch we introduced a brand new TSAM Dashboard, which is powered by Oracle JET. There were some other important enhancements in that patch. So, basically TSAM GA + RP003 makes TSAM a new minor release, internally we call it 12.2.2.1. So, we definitely want to show these latest features.
@Djelibeybi Can I remove the RP as a requirement but still leave it as an option for the users who have rights to obtain the patch?

Djelibeybi: We have to limit the primary `Dockerfile` to things that are available via OTN and other public download sources. You can certainly add a sample of how to apply a rolling patch though.

chrisguoado: OK, that's great!

Djelibeybi:
Also, I just noticed this is under the `OracleTuxedo` directory. Please move it up a level so that `OracleTSAM` is in the root of the repo. The convention is that each product gets its own directory hierarchy.

chrisguoado:
OK. I actually considered about it, and decided to put TSAM inside Tuxedo directory since, TSAM is one of the Tuxedo family products, along with several other products (see [Tuxedo download page](http://www.oracle.com/technetwork/middleware/tuxedo/downloads/index.html)), we used to put them under the Tuxedo brand.

Djelibeybi:
Yeah, but that's not how we've (currently) structured things here in the repo. :) Though, as we get bigger, I may look at restructuring further, but for now, this needs to go up a level so it's easier to discover.

chrisguoado:
@Djelibeybi It's done, please help check again, thanks!

OracleTSAM/samples/xe-db-svc/README.md
```diff
@@ -0,0 +1,42 @@
+Running Oracle TSAM Plus with oracle/database:11.2.0.2-xe image
+=============================
+You can run the TSAM container with the `oracle/database:11.2.0.2-xe` image provided on [GitHub](https://github.com/oracle/docker-images/tree/master/OracleDatabase).
+
+Due to the `oracle/database:11.2.0.2-xe` container creates new database and generates random `sys` user password on startup, a little "hacking" is needed here to make the `sys` password fixed.
```
gvenzl: This should be rephrased. There is no hacking needed to provide a common password for the database container. The automatic generated password was put in place as per Oracle Security Policy. It can be set to whatever is desired by using the `setPassword.sh` script inside the container.

OracleTSAM/samples/xe-db-svc/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Basically this is done by overriding the default command `runOracle.sh` with this [modified one](cmd/runOracle.sh) in the `cmd` folder. The `docker-compose.yml` file to start the containers will look like below:
```
gvenzl: The password is randomly generated as per policy. This was discuss with @Djelibeybi before. I don't think it makes sense to provide a copy of the run script as I can't see how it will stay in sync with the actual image. The default password can be easily overwritten with the `setPassword.sh` script inside the DB container without having to know the generated one. I highly advise to use that script if possible.

chrisguoado: @gvenzl Thank you for the comments. You are right, I also have a strong feeling now I need to rephrase those sentences.

Sorry for letting you confuse on the "hacking" word. I just wanted to express that replacing the database container's entry point command `runOracle.sh` with a modified one on its initialization to force using a designated password looks a little tricky.

The modified version of `runOracle.sh` actually just did a minor change to the original one by calling the `setPassword.sh` at near end of the script to set the SYS password to the one passed in by environment variable `DBA_PASSWD`.

Actually I also realized there could be some sync issue of this modified script with the base one in the future. While considering it's just provided as a sample and it's very easy for users to find out the changes by diffing the two, I think it would be okay to include it.

The intention of this change was to make the instantiation of the TSAM container along with its dependency could be done in one step (the `docker-compose up` command). Otherwise there will be at least 2 steps:
1. Run the database container with `docker run` command. Staring at the output or come back checking the result later until the initialization completes.
2. If a custom SYS password is desired, use `setPassword.sh` to change the password inside the database container.
3. Start the TSAM container and pass in the database connection string and credentials.

I think to eliminate all these efforts for overriding the default container command, or make it possible to use the database image in an automated environment, we may need to do some small enhancement to the database image. Do you think it's feasible for the database container to use a passed in SYS password on its initialization? For example, change line 58 of `runOracle.sh` to something like:
```bash
ORACLE_PWD=${ORACLE_PWD:-`openssl rand -hex 8`}
```
This will use the value of environment variable `ORACLE_PWD` as SYS password if it already exists, or else generate random password.

Djelibeybi: If @gvenzl is OK with this change, you should include it in the pull request. I think we need to hold the merge until this security concern is addressed.

chrisguoado: Yes, I'm waiting for response from @gvenzl

Djelibeybi:
@chrisguoado how about you split the PR into two? One for the base `Dockerfile` for TSAM Plus and its supporting scripts (which could probably be merged) and another one for the samples, which  we can continue to work on with @gvenzl?

If that's OK, go ahead and close this PR and open two new ones.

chrisguoado:
@Djelibeybi How about keeping this PR but remove the two samples `xe-db-svc` and `dashboard-ng` which are currently controversial from my merging branch? This should also make this PR good to merge.

Djelibeybi:
You can do that too, and then submit a new PR with just those samples, so we can continue the discussion in the new PR.

chrisguoado:
OK, I'll let you know when it's ready.

Djelibeybi:
That's great. @gvenzl can you approve these changes, so I can merge this base PR and we can continue the discussion on the samples on a new PR?

OracleTuxedo/OracleTSAM/README.md
```diff
@@ -0,0 +1,174 @@
...
+services:
+  tsam:
+    image: oracle/tsam:12.2.2.1
+    hostname: tsam.docker
+    ports:
+      - 7001/tcp
+      - 22/tcp
+    privileged: true
+    environment:
+      - "DB_CONNSTR=db.box:1521/orcl"
```
chrisguoado: Here the `docker-compose` snippet is provided as a general sample, which is not differentiating the possible Oracle Database types, which could be XE/SE/CDB/PDB.

OracleTuxedo/OracleTSAM/README.md
```diff
@@ -0,0 +1,174 @@
...
+### Running Oracle TSAM Plus with an database Docker container
+If you got an Oracle Database Docker image, you can run it together with the TSAM container to provide the database service. The `docker-compose.yml` file will look like this:
+
+```yaml
+version: "2"
+services:
+  db:
+    image: oracle-database-image
+    hostname: db.box
+    shm_size: 2g
```
chrisguoado: `shm_size` removed.

chrisguoado:
@gvenzl **Anything else to improve?**

OracleTSAM/samples/xe-db-svc/README.md
gvenzl: Yeah, I think that that makes sense. So users have the best of both worlds, either pass on a password or just generate a random one.

Djelibeybi:
Thanks everyone. Merged.

PR #339, opened 2017-04-04
Include known issues section in ReadMe

gvenzl:


OracleDatabase/README.md
```diff
@@ -135,6 +135,9 @@ Another option is to use `docker exec` and run `sqlplus` from within the same co
 
 	docker exec -ti <container name> sqlplus pdbadmin@ORCLPDB1
 
+## Known issues
+* OverlayFS (`overlay`) on CentOS has proven to run into Docker bug #25409. We recommend using `btrfs` or `overlay2` instead. For more details see issue #317.
```
Djelibeybi: It's not OverlayFS causing the problem, it's the `overlay` graph driver for Docker. Both the `overlay` and `overlay2` graph drivers use the same underlying OverlayFS in the kernel.

gvenzl: Ok, can the ordinary user differentiate between OverlayFS and the graph drivers? Maybe we can just say "`overlay` on CentOS has proven..."

Djelibeybi: Perhaps "The [`overlay` storage driver](https://docs.docker.com/engine/userguide/storagedriver/selectadriver/) on CentOS..."

gvenzl: Done!

PR #340, opened 2017-04-05
Add build argument to enter in the command line the WebLogic Admin Se…

mriccell:
…rver name.  Now each Admin Server can have its unique name.


OracleWebLogic/samples/1221-domain/container-scripts/create-wls-domain.py
```diff
@@ -8,13 +8,15 @@
 # Author: bruno.borges@oracle.com
 # ==============================================
 domain_name  = os.environ.get("DOMAIN_NAME", "base_domain")
+admin_name   = os.environ.get("ADMIN_NAME", "AdminSer")
```
brunoborges: @mriccell sure this was supposed to be "AdminSer", not "AdminServer"?

PR #344, opened 2017-04-09
OracleGoldenGate submodule

lonedevel:
These are components for appending the required components for GoldenGate for the Oracle docker image so that you can create a GoldenGate source or target database instance.   Currently its build upon the Oracle 12.1.0.2 database release and GoldenGate 12.2.0.1.1.   I made specific copies of some of the original scripts that Gerald created in the database docker environment to tweak specific settings so that GoldenGate could run and some other environmental requirements to run GoldenGate.

### DIFF
--- a/OracleGoldenGate/Dockerfile
+++ b/OracleGoldenGate/Dockerfile
@@ -44,6 +44,8 @@ ENV ORACLE_BASE=/opt/oracle 					\
     SETUP_LINUX="startDB.sh" 					\
     INSTALL_OGG_FILE="installOGG.sh" 
 
+# Additional environment variables that are populated from first
+# set of environment variables
 ENV INSTALL_DIR=$ORACLE_BASE/install 				\
     LD_LIBRARY_PATH=$ORACLE_HOME/lib:/usr/lib 			\
     CLASSPATH=$ORACLE_HOME/jlib:$ORACLE_HOME/rdbms/jlib		\
@@ -67,12 +69,8 @@ RUN chmod ug+x $INSTALL_DIR/*.sh && \
     sync && \
     $INSTALL_DIR/$SETUP_LINUX_FILE
 
-# Setup rlwrap aliases for use with ggsci and SQL*Plus to enable command line completion.
-# Note for SQL*Plus you can also use sqlcl locally on your host environment.
 USER oracle
-RUN $INSTALL_DIR/$INSTALL_OGG_FILE	 			&& \
-    echo "alias ggsci='rlwrap ggsci'" >> ~/.bashrc              && \
-    echo "alias sqlplus='rlwrap sqlplus'" >> ~/.bashrc
+RUN $INSTALL_DIR/$INSTALL_OGG_FILE
 
 WORKDIR /home/oracle
 
--- a/OracleGoldenGate/Dockerfile
+++ b/OracleGoldenGate/Dockerfile
@@ -1,0 +1,81 @@
+# LICENSE CDDL 1.0 + GPL 2.0
+#
+# Copyright (c) 1982-2017 Oracle and/or its affiliates. All rights reserved.
+#
+# ORACLE DOCKERFILES PROJECT
+# --------------------------
+# This is a Dockerfile for Oracle GoldenGate 12c Release 1.   It depends upon the 
+# Docker image for Oracle Database at: 
+# https://github.com/oracle/docker-images/tree/master/OracleDatabase.
+#
+# REQUIRED FILES TO BUILD THIS IMAGE
+# ----------------------------------
+# (1) fbo_ggs_Linux_x64_shiphome.zip
+#     
+#     Download Oracle GoldenGate 12c Release 1 Enterprise Edition for Linux x64
+#     from http://www.oracle.com/technetwork/middleware/goldengate/downloads/index.html
+#
+# HOW TO BUILD THIS IMAGE
+# -----------------------
+# Put all downloaded files in the same directory as this Dockerfile
+# Run:
+#      $ docker build -t 
+#
+# Pull base image
+# ---------------
+
+FROM oracle/database:12.1.0.2-ee
+
+# Maintainer
+# ----------
+MAINTAINER Rick Michaud<rick.michaud@oracle.com>
+
+# Environment variables required for this build (do NOT change)
+# -------------------------------------------------------------
+ENV ORACLE_BASE=/opt/oracle 					\
+    ORACLE_HOME=/opt/oracle/product/12.1.0.2/dbhome_1 		\
+    OGG_INSTALL_FILE="fbo_ggs_Linux_x64_shiphome.zip" 		\
+    OGG_INSTALL_RSP="oggcore.rsp"				\
+    RUN_FILE="runOracle.sh" 					\
+    START_FILE="startDB.sh" 					\
+    RUN_FILE="runOracle.sh" 					\
+    CREATE_DB_FILE="createDB.sh" 				\
+    SETUP_LINUX_FILE="setupLinuxEnv.sh" 			\
+    SETUP_LINUX="startDB.sh" 					\
+    INSTALL_OGG_FILE="installOGG.sh" 
+
+ENV INSTALL_DIR=$ORACLE_BASE/install 				\
+    LD_LIBRARY_PATH=$ORACLE_HOME/lib:/usr/lib 			\
+    CLASSPATH=$ORACLE_HOME/jlib:$ORACLE_HOME/rdbms/jlib		\
+    OGG_INSTALL_DIR=$ORACLE_BASE/ogg 				
+
+# Setup $PATH properly for environment
+ENV PATH=$ORACLE_HOME/bin:$ORACLE_HOME/OPatch/:/usr/sbin:$PATH 	\
+    PATH=$OGG_INSTALL_DIR:.:$PATH 				\
+    GGHOME=$OGG_INSTALL_DIR
+
+# Copy binaries
+# -------------
+# Copy GoldenGate components to installation directory
+COPY $OGG_INSTALL_FILE $OGG_INSTALL_RSP $SETUP_LINUX_FILE $INSTALL_OGG_BINARIES_FILE $INSTALL_DIR/
+
+# Overlay the database creation run script with the one ammended with GoldenGate requirements
+COPY $RUN_FILE $CREATE_DB_FILE $ORACLE_BASE/
+
+USER root
+RUN chmod ug+x $INSTALL_DIR/*.sh && \
+    sync && \
+    $INSTALL_DIR/$SETUP_LINUX_FILE
+
+# Setup rlwrap aliases for use with ggsci and SQL*Plus to enable command line completion.
+# Note for SQL*Plus you can also use sqlcl locally on your host environment.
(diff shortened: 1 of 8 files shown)


OracleGoldenGate/setupLinuxEnv.sh
```diff
@@ -0,0 +1,30 @@
...
+# Setup environment for GoldenGate.   Also install rlwrap, vim and Java 1.8 for other requirements.
+mkdir -p $OGG_INSTALL_DIR && \
+chown -R oracle:oinstall $INSTALL_DIR && \
+chown -R oracle:oinstall $OGG_INSTALL_DIR && \
+chmod ug+x $INSTALL_DIR && \
+chmod ug+x $OGG_INSTALL_DIR && \
+wget http://www.mirrorservice.org/sites/dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
+rpm -Uvh epel-release-latest-7.noarch.rpm && \
+yum -y -q install rlwrap && \
+wget --no-cookies --no-check-certificate --header "Cookie: gpw_e24=http%3A%2F%2Fwww.oracle.com%2F; oraclelicense=accept-securebackup-cookie" "http://download.oracle.com/otn-pub/java/jdk/8u60-b27/jdk-8u60-linux-x64.rpm" && \
```
Djelibeybi: Please do not bypass the licence acceptance mechanism to download Oracle JDK. You will need to document the JDK requirement and have users download it manually. You'll then need to copy the RPM Into the image via the Dockerfile during build.

OracleGoldenGate/setupLinuxEnv.sh
```diff
@@ -0,0 +1,30 @@
...
+# ------------------------------------------------------------
+
+# Setup environment for GoldenGate.   Also install rlwrap, vim and Java 1.8 for other requirements.
+mkdir -p $OGG_INSTALL_DIR && \
+chown -R oracle:oinstall $INSTALL_DIR && \
+chown -R oracle:oinstall $OGG_INSTALL_DIR && \
+chmod ug+x $INSTALL_DIR && \
+chmod ug+x $OGG_INSTALL_DIR && \
+wget http://www.mirrorservice.org/sites/dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
+rpm -Uvh epel-release-latest-7.noarch.rpm && \
```
Djelibeybi: Please do not install EPEL as this is unsupported by Oracle and will result in an unsupported image. If there is a package missing from Oracle Linux, please raise this internally.

lonedevel:
Ok I'll pull out the Java and EPEL/rlwrap components .    Java isn't really required for this image, but I was trying to stage it for another subproject and rewrap is more of a convenience for GGSCI command line tool.

brunoborges:
@lonedevel, please email github_info_ww_grp@oracle.com requesting for information on how to contribute as employee to Oracle projects on GitHub

lonedevel:
Ok will do.   Is there some sort of approval process?

lonedevel:
Made requested changes in latest commit.

OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+# HOW TO BUILD THIS IMAGE
+# -----------------------
+# Put all downloaded files in the same directory as this Dockerfile
+# Run:
+#      $ docker build -t 
+#
+# Pull base image
+# ---------------
+
+FROM oracle/database:12.1.0.2-ee
```
gvenzl: The image should be based on the latest 12.2 EE version.

OracleGoldenGate/createDB.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
gvenzl: This file should just be reused from the base container and not being duplicated. You can just call `createDB.sh` given that your image is based on the database image.

OracleGoldenGate/runOracle.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
```
gvenzl: This file should be reused from the base container and not being duplicated. `runOracle.sh` is available inside your container as you have based the image on the database image.

OracleGoldenGate/setupLinuxEnv.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# LICENSE CDDL 1.0 + GPL 2.0
+#
+# Copyright (c) 1982-2016 Oracle and/or its affiliates. All rights reserved.
+#
+# Since: December, 2016
+# Author: gerald.venzl@oracle.com
```
gvenzl: The header of this file needs changing as it has nothing to do anymore with database installation. And I also don't want to get emails about GoldenGate setups. :)

OracleGoldenGate/Dockerfile
lonedevel: Current GoldenGate 12.3 is not GA which supports DB 12.2EE yet so I'm just waiting for that to be available before I provide support for it for capture and apply.

OracleGoldenGate/createDB.sh
lonedevel: So I couldn't find an elegant way to integrate the required changes for GoldenGate upon database creation to enable archive logging,  supplemental logging and create the necessary GoldenGate user with the necessary privileges.   I had considered doing it after instance creation but I'd have to somehow wait until the createDB is complete before I run the additional steps outlined below.  I'm open to suggestions if you have any thoughts...

I added these to the bottom of that the createDB.sh:   

`# Modified by Rick Michaud for GoldenGate purposes
# Setting up database logging for GoldenGate
echo "Configuring database logging for GoldenGate..."
sqlplus / as sysdba << EOF
   set echo on
   set timing on
   SHUTDOWN IMMEDIATE;
   STARTUP MOUNT;
   ALTER DATABASE ARCHIVELOG;
   ALTER DATABASE OPEN;
   ALTER SYSTEM SWITCH LOGFILE;
   ALTER SESSION SET CONTAINER=$ORACLE_PDB;
   ALTER PLUGGABLE DATABASE ADD SUPPLEMENTAL LOG DATA;
   exit;
EOF

# Create GoldenGate User for Source Extract
echo "Creating GoldenGate user and granting appropriate privileges..."
sqlplus / as sysdba << EOF
   set echo on
   set timing on
   CREATE USER C##ggadmin IDENTIFIED BY ggadmin DEFAULT TABLESPACE USERS TEMPORARY TABLESPACE TEMP;
   EXEC dbms_goldengate_auth.grant_admin_privilege('C##ggadmin', CONTAINER=>'ALL');
   GRANT DBA TO C##ggadmin CONTAINER=ALL;
   exit;
EOF
`

OracleGoldenGate/runOracle.sh
lonedevel: Ok I'll reuse the existing one from the base container.

OracleGoldenGate/setupLinuxEnv.sh
lonedevel: Ok will make the changes.

OracleGoldenGate/Dockerfile
gvenzl: Ok, thanks for clarification, @lonedevel! I just wanted to make sure!

OracleGoldenGate/createDB.sh
gvenzl: I think what you should do is call `createDb.sh` and afterwards have your own script, e.g. "setupGoldenGate.sh" that will perform those steps that you have outlined above. The run script will detect whether it has to create the DB or it's already there so from there I think you are fine. But I did spot one thing with the `runOracle.sh` script. The hard coded `tail -f` at the end. I will have to enhance the `runOracle.sh` script to allow for batch mode, i.e. not wait and tail on the alert log but just finish. That would allow you to just execute `createDb.sh`, `setupGoldenGate.sh` and `runOracle.sh` and then proceed with whatever is appropriate for your container.

OracleGoldenGate/setupLinuxEnv.sh
```diff
@@ -0,0 +1,23 @@
...
+# Adjust file permissions, go to /opt/oracle as user 'oracle' to proceed with Oracle installation
+# ------------------------------------------------------------
+
+# Setup environment for GoldenGate.   
+mkdir -p $OGG_INSTALL_DIR && \
+chown -R oracle:oinstall $INSTALL_DIR && \
+chown -R oracle:oinstall $OGG_INSTALL_DIR && \
+chmod ug+x $INSTALL_DIR && \
+chmod ug+x $OGG_INSTALL_DIR && \
+yum -y -q install vim-enhanced && \
```
Djelibeybi: Why are you installing `vim-enhanced` into a container? There should be no reason for you to edit files in a running container. Any changes you make should be done either during build or in an external volume.

lonedevel: So the GoldenGate command line tool, ggsci, uses a Unix editor when setting up configuration for replication data flows.   Each processes configuration is done through this process to create a parameter file.   It could be mapped through an external volume, however ggsci tries to load an external editor (like vi, emacs or nano) when you try to alter the settings in one of these parameter files.    These are usually kept on the same host environment where the source or target database resides.    See: [GGSCI](http://docs.oracle.com/goldengate/c1221/gg-winux/GWUAD/getting-started-oracle-goldengate-process-interfaces.htm#GUID-ED4CCFF1-CE02-4E73-8D9F-790A12B55A7E )for deeper details.

Djelibeybi: But is this something you would do inside a running container? I don't have experience with GoldenGate (and I'm sure it shows!), but best practices suggest that you shouldn't need to modify a running container manually. This would break things like automation and orchestration.

lonedevel: Actually it is something a user would do in a running container such as  Oracle database container where they can set up the GoldenGate configuration while the database is running.   The configuration through GoldenGate parameter files  is the "development" process used to build a replication data flow configuration and the built-in editor functions as the IDE.   This allows the image to be generic, but having the ability to spawn multiple containers to represent a GoldenGate pattern.  So for example there could be a source container with GoldenGate + the Oracle database and a target container with GoldenGate + the Oracle database.     Or there could be GoldenGate running on an Oracle database container streaming data to something like the WebLogic JMS container or to a Confluent Kafka container.     Basically, you have to think of it like a development environment.   

Now I can certainly access the GoldenGate configuration externally using the hosts editor or the GoldenGate Studio UI, or I can put these tools in their own image and have them run under a container.    For now I'm just starting with little steps and still need the editor within the container.   I get it that this adds extract "weight" to the container and the intent is to be be simple and light.

Djelibeybi: In which case, I'd suggest installing `vim-minimal` instead, which should be sufficient, but won't pull in 30+ Perl packages like `vim-enhanced` does.

Djelibeybi: Also, I would strongly encourage you to at least document how to access the GoldenGate configuration externally or connect via GoldenGate Studio into a running container.

Djelibeybi:
Closing this in favour of PR #431.

PR #365, opened 2017-04-27
Add Dockerfile to create WLS infrastructure image

mriccell:


Djelibeybi:
@mriccell can you perhaps update the README for the WebLogic images to provide some detail on the differences between the `generic`, `developer` and now `infrastructure` images are?

PR #380, opened 2017-05-16
Adding VOLUME for persisting kvroot

nirmay:
please review.


brunoborges:
Can you please explain this change and its use case?

Djelibeybi:
I'm not sure why this change needs explaining, @brunoborges. :) It's obviously intended to store the NoSQL data in a persistent volume so that the data can persist a container destruction/creation cycle.

brunoborges:
@Djelibeybi it was only for the record... :P

Djelibeybi:
Though, I suspect this is not the most efficient way to add `VOLUME` support, because `KVHOME` is versioned. It would probably be better to change `KVHOME` to a more generic name so that when NoSQL is upgraded, the data is persisted.

I'm going to take a look at this later and perhaps submit another PR to make this a bit clearer.

PR #385, opened 2017-05-16
Removed references to Server JRE 7u80

aureliogrb:
7u80 is several years old and is no longer maintained.  It is missing many security vulnerability fixes.  The download page that this document referred to clearly states this should not be used in product.


Djelibeybi:
I've also removed the `java-7` and `windows-java-7` directories completely.

PR #389, opened 2017-05-18
Issue-388 - Added support to execute scripts after database is up

jcustenborder:
Added support to launch scripts during startup.

### DIFF
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -140,6 +140,17 @@ Another option is to use `docker exec` and run `sqlplus` from within the same co
 
 	docker exec -ti <container name> sqlplus pdbadmin@ORCLPDB1
 
+### Running scripts on startup
+The docker images can be configured to run scripts on startup. Currently `sh` and `sql` extensions are supported.  You can either 
+mount the volume `/docker-entrypoint-initdb.d` or extend the image to include scripts in this directory. After the image is started
+the scripts in `/docker-entrypoint-initdb.d` will be executed against the image. SQL scripts will be executed as sysdba, shell scripts 
+will be executed as the current user. To ensure proper order it is recommended to prefix your scripts with a number. For example 
+`01_users.sql`, `02_permissions.sql`, etc. The example below mounts the local directory scripts to `/docker-entrypoint-initdb.d`.
+
+    docker run --rm -ti -v scripts:/docker-entrypoint-initdb.d oracle/database:12.2.0.1-ee sqlplus pdbadmin/<yourpassword>@//<db-container-ip>:1521/ORCLPDB1
+    
+
+
 ## Known issues
 * The [`overlay` storage driver](https://docs.docker.com/engine/userguide/storagedriver/selectadriver/) on CentOS has proven to run into Docker bug #25409. We recommend using `btrfs` or `overlay2` instead. For more details see issue #317.
 
--- a/OracleDatabase/dockerfiles/11.2.0.2/runOracle.sh
+++ b/OracleDatabase/dockerfiles/11.2.0.2/runOracle.sh
@@ -132,4 +132,14 @@ echo "#########################"
 
 tail -f $ORACLE_BASE/diag/rdbms/*/*/trace/alert*.log &
 childPID=$!
+
+for f in /docker-entrypoint-initdb.d/*; do
+    case "$f" in
+        *.sh)     echo "$0: running $f"; . "$f" ;;
+        *.sql)    echo "$0: running $f"; echo "exit" | $ORACLE_HOME/bin/sqlplus -s "/ as sysdba" @"$f"; echo ;;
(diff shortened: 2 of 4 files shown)


Djelibeybi:
@brunoborges can you follow-up with OCA requirements? I like this enhancement. :)

brunoborges:
@jcustenborder can you please [sign this](http://www.oracle.com/technetwork/community/oca-486395.html) and when you email to Oracle, please cc me (bruno dot borges at oracle dot com)

jcustenborder:
@brunoborges Hah my scanner didn't want to work the other day and I forgot. I have it signed sitting in the scanner. I set a reminder to do it when I get home tonight. Are there any other changes that are required?

brunoborges:
I don't think so.

nicolimo86:
Shouldn't be documented this functionality? I mean in the README.

NMichas:
We'd love to see this being merged!

jcustenborder:
@brunoborges Sorry about the delay. I've been traveling then my scanner was having issues. I emailed you something.

jcustenborder:
Sorry for the delay on this. I was told I need to change one thing on the
scanned paperwork I sent Oracle. Unfortunately I'm on holiday until the
12th with limited access to the internet.

gvenzl:
No rush, enjoy your holiday! Once you are back we can take a look at this which is convenient for me as well because I too am traveling at the moment, albeit not for holiday :(

brunoborges:
@jcustenborder your OCA was accepted. We are ready to take this in, but can you check @gvenzl's comments?

jcustenborder:
@brunoborges Just saw this. Will do now.

jcustenborder:
@gvenzl Sorry about missing your comments. Can you let me know if this is what you are looking for?

jcustenborder:
@gvenzl Comming back around on this one. Are there any other changes you need from me?

gvenzl:
Hi @jcustenborder, thanks a lot for your contribution. I have tested your changes and they work fine.

Just minor comments:

* After about 5 times having typos typing /docker-entrypoint-initdb.d I thought it may be better to keep it in sync with the other volume `/opt/oracle/oradata` and just use `/opt/oracle/scripts` instead. I think it just will be easier for folks to remember, please let me know your thoughts.
* I amended the example that you have in the ReadMe slightly as Docker does require an absolute path to a host mounted volume and because the password at this stage is auto generated and not known to the user if not specified directly.
* The XE version did actually not work due to the way how XE is run (under the root user rather than oracle)

Furthermore, I have added a hopefully simple to follow sample.

You can have a look at the changes here: https://github.com/oracle/docker-images/compare/389-startupscripts

If you agree with the changes I will file a PR containing yours and those and merge those back into the master.

Please let m know your thoughts.

jcustenborder:
> After about 5 times having typos typing /docker-entrypoint-initdb.d I thought it may be better to keep it in sync with the other volume /opt/oracle/oradata and just use /opt/oracle/scripts instead. I think it just will be easier for folks to remember, please let me know your thoughts.

I used /docker-entrypoint-initdb.d because this is a standard location. MySQL and Postgres use this location. I was trying to keep things consistent.  

> The XE version did actually not work due to the way how XE is run (under the root user rather than oracle)

Thanks for catching this. Are there any changes you recommend?

gvenzl:
I see, in that case it may makes sense to either keep that or perhaps just look in both directories. I'm also starting to think whether we don't need to separate setup and startup scripts. Right now, there is no way to run a script only at the setup time of the database but not when restarting it unless you put some logic into your scripts.

Just a slight change needed using the `su` command, check https://github.com/oracle/docker-images/compare/389-startupscripts#diff-c7ba2f7176dd4e0d1f80615d2a83067bR142

Djelibeybi:
My (unrequested/unwarranted) opinion: I prefer `/docker-entrypoint-initdb.d` for the same reason @jcustenborder mentioned: it's already in use by both MySQL and PostgreSQL.

And I think separating the scripts into setup (i.e. first-run when the DB is initially created) and startup makes excellent sense too.

gvenzl:
Indeed, but at that point the path will become inconsistent with MySQL and Postgres anyway. Probably still best to keep both and search for both, e.g. if `/docker-entrypoint-initdb.d` does not exist/is empty, search for `/opt/oracle/scripts` or something like that. Best of both worlds.

jcustenborder:
@gvenzl Did you want to add this or did you want me to do it?

gvenzl:
@jcustenborder, I'm happy to make the changes if that is fine with you.

jcustenborder:
Works for me!

gvenzl:
@jcustenborder, I have filed a new PR #471 which incorporates your contributions as well as the changes discussed.

gvenzl:
Closing this PR in favor of #471. Thanks a lot for your contribution, @jcustenborder! Very much appreciated!

PR #393, opened 2017-05-20
Dockerfile.template fixed

mikarinneoracle:


ContainerCloud/images/rolling-router-sticky-sessions/Dockerfile.template
```diff
@@ -71,9 +71,11 @@ ADD nginx-files/00-upstream-placeholder \
     /etc/nginx/sites-enabled/00-upstream
 
 ADD nginx.sh /etc/sv/nginx/run
+RUN chmod +x /etc/sv/nginx/run
```
karthequian: @mikarinneoracle in the future, can we make this in a single run block rather than multiples?

mikarinneoracle: Good suggestion, thanks. Will do!

PR #425, opened 2017-06-27
Docker related scripts for OHS 12.2.1.2.0

prkishor:


Djelibeybi:
FYI: I'm offline for the rest of the week (mostly), so I'll only be able to review this PR next week.

prkishor:
I am highlighting the changes broadly. they are as follows:
1.Change in the directory structure inside the OracleHTTPServer folder. The 'samples' folder has been removed and the scripts inside 'container-scripts' have been moved inside 12.2.1.2.0/container-scripts with some modifications.
2.References to 12.2.1 have been replaced with 12.2.1.2.0 (renaming of folder and references in messages/README file).
3.The Dockerfile now creates a standalone domain(thereby creating OHS instance) and starts the OHS instance as well instead of just creating a standalone OHS installation.
4.Added SIGTERM and SIGKILL handler in startNMandOHS.sh
5.Created two new files createDomain.sh(to be called from Dockerfile) and stop-ohs.py(to handle SIGTERM request)
6.Modifications in main Dockerfile(the only one now) for eg-- No JAVA Home is being provided during execution of OHS installer as it is not required.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,60 +4,63 @@
...
 # For example
 # 
-#     $ docker build --force-rm=true --no-cache=true --rm=true -t oracle/ohs:12.2.1-sa .
+#     $ docker build --force-rm=true --no-cache=true --rm=true -t oracle/ohs:12.2.1.2.0-sa .
 #
-# Pull Oracle Linux 7 base image
+# Pull serverjre:8 base image
 # -----------------------------------
-FROM oracle/serverjre:8
+FROM container-registry.oracle.com/java/serverjre:8
```
Djelibeybi: Do not specify the container-registry URL here. Customers can build `java/serverjre:8` locally or could use the Docker Store version too.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -67,21 +70,39 @@ RUN yum install -y \
         gcc gcc-c++ glibc glibc-devel libaio libaio-devel libgcc \
 	libstdc++ libstdc++-devel ksh make ocfs2-tools sysstat && \
```
Djelibeybi: Do you really need `ocfs2-tools`? There is no way to mount an OCFS2 filesystem from inside a container.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -67,21 +70,39 @@ RUN yum install -y \
...
+    mkdir /u01/oracle/container-scripts/ /u01/oracle/logs 
+
+# Copy required files to build this image
+# ------------------------------------------------------
+COPY container-scripts/* /u01/oracle/container-scripts/
+
+# Configuration of OHS domain
+# Copy static html for sanity check
+#---------------------------------------------------
+RUN echo "export PATH=$PATH:$ORACLE_HOME/oracle_common/common/bin:$ORACLE_HOME/user_projects/domains/ohsDomain/bin" >> /u01/oracle/.bashrc
```
Djelibeybi: I'm not entire sure what the worth of this is: a Docker container rarely runs `bash` so modifying the `.bashrc` file for the `oracle` user would only trigger if someone manually exec'd `bash` in a running container.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -67,21 +70,39 @@ RUN yum install -y \
         gcc gcc-c++ glibc glibc-devel libaio libaio-devel libgcc \
 	libstdc++ libstdc++-devel ksh make ocfs2-tools sysstat && \
     yum clean all && \
-    rm -rf /var/cache/yum/* && \
     chmod a+xr /u01 && \
     useradd -b /u01 -m -s /bin/bash oracle && \
     echo oracle:oracle | chpasswd && \
```
Djelibeybi: You are hardcoding a password for the `oracle` user which is against security policy. I don't see any reason why you're setting a password anyway. There is no way to remotely login as the `oracle` user.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
Djelibeybi: Also, why do you need `ksh`? I would've assumed `bash` would be sufficient.

prkishor:
Hi Avi,

It is listed in the doc.

http://docs.oracle.com/html/E77908_01/toc.htm

THanks,

Prabhat

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -68,11 +68,10 @@ COPY $OHS_PKG install.file oraInst.loc /u01/
 RUN yum install -y \
         unzip tar nfs-utils binutils compat-libcap1 compat-libstdc++ \
```
Djelibeybi: I just noticed `nfs-tools` here and I'm still not convinced about why you need `ksh`. Just pointing to the documentation is not sufficient. You need to evaluate all the requirements to determine if they're still valid, because anything you add will make the image larger.

Djelibeybi:
Excellent work, thanks.

rdas76:
1) README - It's not "Oracle JDK 8 (Server)", it's Oracle Server JRE 8. Pls change the name to Oracle Server JRE 8.

2) Volume usages -
 a) You are creating a Docker volume and using the standard Docker volume location /var/lib/docker/volume. In our [Best Practices](https://orahub.oraclecorp.com/fmw-cert-dev/fmw-docker-images/tree/master/bestpracticeReferences/OracleSOA-reference) we have used non standard volume location. We do not execute separate command to create a volume. I would recommend to follow the same approach for consistencies.

 b)  We have put the entire user_projects in the volume but in case of OHS image this is not happening. The purpose of putting the user_projects in the volume is to separate and persist the configuration data. This will also help in readability and debugging of any failures. 

3) Stop/Restart Scripts - There are scripts for stop and restart of servers but README does not have any description as to how to use them.

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
+If we want to start the OHS container without specifying any configuration for mod_weblogic:
+1. To start the OHS container with above oracle/ohs:12.2.1.2.0-sa image, run the following command:
 
-         Eg:$ docker volume create --name volume
+         docker run -it --name ohs -p 7777:7777 oracle/ohs:12.2.1.2.0-sa
 
-_This volume will be created in "/var/lib/docker" directory or the location where "/var/lib/docker" points to._
 
+If we want to start the OHS container with some pre-specified mod_weblogic configuration:
+1. Depending on your weblogic environment , create a **custom_mod_wl_ohs.conf** file by referring to container-scripts/mod_wl_ohs.conf.sample and section 2.4 @ [OHS 12c Documentation](http://docs.oracle.com/middleware/12212/webtier/develop-plugin/oracle.htm#PLGWL553)
```
Djelibeybi: Remove the space before the comma, please.

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
+You must first download the Oracle Server JRE binary and drop in folder `../OracleJava/java-8` and build that image. For more information, visit the [OracleJava](../OracleJava) folder's [README](../OracleJava/README.md) file.
 
-### How to Build OHS domain Image
-Go to folder **samples/1221-OHS-domain**
+        $ cd ../OracleJava/java-8
+        $ sh build.sh
 
-Run the following command:
+### Building OHS Docker Image
+IMPORTANT: You have to download the OHS binary and put it in place (see .download files inside dockerfiles/).
```
Djelibeybi: I would specify the URL and OHS binary to download instead of relying on the contents of the `.download` file. We've had users in the past who've not completely understood this concept, so the more documentation on what to download and from where is better.

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
-         Eg:$ docker volume create --name volume
+         docker run -it --name ohs -p 7777:7777 oracle/ohs:12.2.1.2.0-sa
 
-_This volume will be created in "/var/lib/docker" directory or the location where "/var/lib/docker" points to._
 
+If we want to start the OHS container with some pre-specified mod_weblogic configuration:
+1. Depending on your weblogic environment , create a **custom_mod_wl_ohs.conf** file by referring to container-scripts/mod_wl_ohs.conf.sample and section 2.4 @ [OHS 12c Documentation](http://docs.oracle.com/middleware/12212/webtier/develop-plugin/oracle.htm#PLGWL553)
 
-1. Depending on your weblogic environment , create a **custom_mod_wl_ohs.conf** file by referring to container-scripts/mod_wl_ohs.conf.sample and section 2.4 @ [OHS 12c Documentation](http://docs.oracle.com/middleware/1221/webtier/develop-plugin/oracle.htm#PLGWL553)
+2. Place the custom_mod_wl_ohs.conf file in a directory in the host say,"/scratch/DockerVoulme/OHSVolume" and then mount this directory into the container as a volume.
```
Djelibeybi: Add a space after the comma, please.

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
 
-2. Place the custom_mod_wl_ohs.conf file in docker data volume directory . e.g /var/lib/docker/volume
-
-3. To start the OHS Container with above sampleohs:12.2.1 image , run command from docker voume directory
+3. To start the OHS Container with above oracle/ohs:12.2.1.2.0-sa image , run command from the directory which ha sbeen mounted as a volume.
```
Djelibeybi: Remove the space before the comma, please.

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
 
          For e.g
-         $ cd /var/lib/docker/volume
-         $ docker run -v `pwd`:/volume -w /volume -d --name ohs -p 7777:7777  sampleohs:12.2.1 configureWLSProxyPlugin.sh
+         $ cd /scratch/DockerVolume/OHSVolume
+         $ docker run -v `pwd`:/volume -w /volume -d --name ohs -p 7777:7777  oracle/ohs:12.2.1.2.0-sa configureWLSProxyPlugin.sh
```
Djelibeybi: You shouldn't need to specify `configureWLSProxyPlugin.sh`. Rather, your `CMD` script should automatically detect if a custom configuration has been provided and fire the script itself.

OracleHTTPServer/README.md
```diff
@@ -95,6 +60,31 @@ _This volume will be created in "/var/lib/docker" directory or the location wher
 
 ######Later you can login to running container and configure Weblogic Server proxy plugin file and run *restartOHS.sh* script.
```
Djelibeybi: You should never need to login to a running container. This should all be handled internally.

OracleHTTPServer/README.md
```diff
@@ -95,6 +60,31 @@ _This volume will be created in "/var/lib/docker" directory or the location wher
...
 
+### Stopping and re-starting OHS instance
+To stop the OHS instance, execute the following steps:
+1. Exec into the container and execute /bin/bash (Assuming the name of container is 'ohs')
```
Djelibeybi: You should never require a user to `exec` into a running container. If you want to stop the OHS instance, you should write handlers for `docker stop` instead.

OracleHTTPServer/README.md
```diff
@@ -95,6 +60,31 @@ _This volume will be created in "/var/lib/docker" directory or the location wher
...
+2. Navigate to /u01/oracle/ohssa/oracle_common/common/bin
+
+      cd /u01/oracle/ohssa/oracle_common/common/bin
+
+3. Execute stop-ohs.py 
+ 
+      ./wlst.sh /u01/oracle/container-scripts/stop-ohs.py
+
+
+To re-start the OHS instance, execute the following steps:
```
Djelibeybi: As above, you should never require a user to `exec` into the container. To stop and restart the process, it should be done using `docker stop` and `docker start`.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,84 +4,97 @@
...
+# Install OL7 required packages. Refer FMW 12.2.1.2 System requirements guide for complete list of packages
 # Change the open file limits and kernel parameters that need changing
 # Setup filesystem and oracle user
 #---------------------------------------------------------------------------------------------------------
 # Proxy needs to be set before running yum in /etc/yum.conf
 RUN yum install -y \
-        unzip tar nfs-utils binutils compat-libcap1 compat-libstdc++ \
-        gcc gcc-c++ glibc glibc-devel libaio libaio-devel libgcc \
-	libstdc++ libstdc++-devel ksh make ocfs2-tools sysstat && \
+        unzip libaio sysstat && \
```
Djelibeybi: This yum command can now be a single line for better readability.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,84 +4,97 @@
...
     yum clean all && \
-    rm -rf /var/cache/yum/* && \
     chmod a+xr /u01 && \
     useradd -b /u01 -m -s /bin/bash oracle && \
-    echo oracle:oracle | chpasswd && \
-    mkdir /u01/oracle/.inventory
+    mkdir /u01/oracle/.inventory && \
+    chown oracle:oracle -R /u01 
 
+# Install and configure Oracle JDK
```
Djelibeybi: The Oracle JDK is already installed and configured, so this comment is no longer required.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,84 +4,97 @@
...
 # Go to /u01 as user 'oracle' to proceed with OHS installation
 #--------------------------------------------------------
-RUN unzip -q -d /u01 /u01/$OHS_PKG && \
-    chown oracle:oracle -R /u01 && \
-    su -c "/u01/$OHS_BIN -ignoreSysPrereqs -novalidation -silent -responseFile /u01/install.file -invPtrLoc /u01/oraInst.loc -jreLoc $JAVA_HOME ORACLE_HOME=$ORACLE_HOME" - oracle && \
-    rm /u01/$OHS_PKG /u01/$OHS_BIN /u01/fmw_*_readme.htm /u01/oraInst.loc /u01/install.file
-
 USER oracle
-WORKDIR /u01/oracle/
+RUN cd /u01 && unzip -q /u01/$OHS_PKG && cd - && \
```
Djelibeybi: You have Java installed, so you can drop the requirement to install `unzip` and run `java xf` instead. This will reduce the size of the image too.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
Djelibeybi: Also, as you can use `java` to unzip the archives, you can probably drop the requirement to install  `unzip` too.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,84 +4,97 @@
...
+EXPOSE $OHS_SSL_PORT $OHS_LISTEN_PORT $NM_PORT
+
+# Final setup
+WORKDIR /u01/oracle/container-scripts
+
+# Create Doamin, OHS instance and start the  OHS instance
+RUN sh createDomain.sh 
+
+# Define default command to start bash
+CMD ["startNMandOHS.sh"]
```
Djelibeybi: This script should automatically detect if a custom configuration has been provided and do the necessary configuration before starting NM and OHS.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,84 +4,97 @@
...
+# ------------------------------------------------------
+COPY container-scripts/* /u01/oracle/container-scripts/
+
+#Expose all Ports
+EXPOSE $OHS_SSL_PORT $OHS_LISTEN_PORT $NM_PORT
+
+# Final setup
+WORKDIR /u01/oracle/container-scripts
+
+# Create Doamin, OHS instance and start the  OHS instance
```
Djelibeybi: Typo: should be Domain.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
```diff
@@ -4,84 +4,97 @@
...
+RUN sh createDomain.sh
```
Djelibeybi: Do you need to do this during build? Can this be shifted into the `CMD` script instead so that the domain is created at run time? This would also allow you to have different configuration of the domain across individual containers without having to rebuild the image.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
Djelibeybi: If you implement the `createDomain.sh` functionality, this script should include that as well.

OracleHTTPServer/README.md
```diff
@@ -41,7 +41,7 @@ If we want to start the OHS container with some pre-specified mod_weblogic confi
 
          For e.g
          $ cd /scratch/DockerVolume/OHSVolume
-         $ docker run -v `pwd`:/volume -w /volume -d --name ohs -p 7777:7777  oracle/ohs:12.2.1.2.0-sa configureWLSProxyPlugin.sh
+         $ docker run -v `pwd`:/u01/oracle/ohssa/user_projects -w /u01/oracle/ohssa/user_projects -d --name ohs -p 7777:7777  oracle/ohs:12.2.1.2.0-sa configureWLSProxyPlugin.sh
```
Djelibeybi: I'd caution against using `pwd` here. Instead of `cd`'ing into `/scratch/DockerVolume/OHSVolume` first, just specify it directly so that the user's current directory is not important:

```
$ docker run -v /scratch/DockerVolume/OHSVolume:/u01/oracle/ohssa/user_projects -w /u01/oracle/ohssa/user_projects -d --name ohs -p 7777:7777  oracle/ohs:12.2.1.2.0-sa
```

OracleHTTPServer/README.md
mriccell: Typo "ha sbeen"

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
-The best way to create your own, or extend domains is by using WebLogic Scripting Tool. You can find an example of a WLST script to create domains at create-sa-ohs-domain.py. You may want to tune this script with your own setup . You can also extend images and override an existing domain, or create a new one with WLST.
-
-## Building a sample Docker Image of a OHS Domain
-To try a sample of a OHS standalone image with a domain configured, follow the steps below:
-
-Make sure you have **oracle/ohs:12.2.1-sa** image built. If not go into **dockerfiles/12.2.1** and call:
-
-        $ sh buildDockerImage.sh -v 12.2.1
+### Building Oracle JDK (Server JRE) base image
+You must first download the Oracle Server JRE binary and drop in folder `../OracleJava/java-8` and build that image. For more information, visit the [OracleJava](../OracleJava) folder's [README](../OracleJava/README.md) file.
```
mriccell: You can also pull the Oracle Server JRE 8 image from [Oracle Container Registry](https://container-registry.oracle.com) or the [Docker Store](https://store.docker.com).

OracleHTTPServer/README.md
```diff
@@ -1,82 +1,47 @@
...
 
-### How to Build OHS domain Image
-Go to folder **samples/1221-OHS-domain**
+        $ cd ../OracleJava/java-8
+        $ sh build.sh
```
mriccell: Please add : You can also pull the Oracle Server JRE 8 image from Oracle Container Registry or the Docker Store.

OracleHTTPServer/README.md
mriccell: Can you provide an example of the command to mount the custom_mod_wl_ohs.conf file into the container volume?

OracleHTTPServer/README.md
mriccell: Please add the URLs for container register and Docker Store

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
brunoborges: Java cannot be used to unzip archives that contain shell scripts. Unfortunately, JAR does not restores unix file permissions, and that may ruin everything. Leave `unzip` in if you are extracting something that has executable files.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
brunoborges: See previous comment. Keep unzip because `jar xf` does not restore file permissions.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
Djelibeybi: That's good to know! Thanks @brunoborges for pointing that out. `unzip` it is.

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
prkishor: This can be done at run time. But, I am not understanding the advantage pointed out by you.
"This would also allow you to have different configuration of the domain across individual containers without having to rebuild the image."

OracleHTTPServer/dockerfiles/12.2.1.2.0/Dockerfile
prkishor: Could you please explain your comment"This script should automatically detect if a custom configuration has been provided and do the necessary configuration before starting NM and OHS."?

prkishor: I am not able to understand what detecting if a custom configuration has been provided is.

OracleHTTPServer/README.md
prkishor: I have defined a SIGTERM handler in my code. That invokes stop-ohs.py when a SIGTERM is received by the container. I will remove this description from README.md

OracleHTTPServer/README.md
prkishor: How do  I handle the restart case? For stopping, I have set a handler for SIGTERM, for restart which signal I should target?

PR #427, opened 2017-06-29
Fix Permission and ADMIN_PASSWORD issue for 12212-domain

bluezd:
There are two commits that apply to be merged:
- 12212-domain: Make user setting Weblogic password successfully
   While specifying the "--build-arg ADMIN_PASSWORD=welcome1" option for building image the 
   varible ADMIN_PASSWORD defined in Dockerfile will never be set due to Docker known 
   mechanism block this behavior. It causes the image will not be built successfully, so this patch 
   intends to fix this issue.

- 12212-domain: Fix the permission issue while building image
   This patch intends to fix the following failure while building image:
   "/bin/sh: /u01/oracle/wlst: Permission denied"

Could you please review those two patches ? Any suggestions are appreciated.

### DIFF
--- a/OracleWebLogic/samples/12212-domain/Dockerfile
+++ b/OracleWebLogic/samples/12212-domain/Dockerfile
@@ -25,7 +25,7 @@ MAINTAINER Monica Riccelli <monica.riccelli@oracle.com>
 
 # WLS Configuration (editable during build time)
 # ------------------------------
-ARG ADMIN_PASSWORD
+ARG ADMIN_PASS
 ARG DOMAIN_NAME
 ARG ADMIN_PORT
 ARG CLUSTER_NAME
@@ -35,6 +35,7 @@ ARG PRODUCTION_MODE
 # WLS Configuration (editable during runtime)
 # ---------------------------
 ENV ADMIN_HOST="wlsadmin" \
+    ADMIN_PASSWORD="${ADMIN_PASS:-welcome1}" \
     NM_PORT="5556" \
     MS_PORT="7001" \
     DEBUG_PORT="8453" \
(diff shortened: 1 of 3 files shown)


mriccell:
I tested these changes and the ARG ADMIN_PASS  when passed in the docker build command is not evaluated in the statement  ENV ADMIN_PASSWORD="${ADMIN_PASS:-welcome1}". Even when set it it with --build-arg ADMIN_PASS=-welcome3, the password is always set to welcome1.  When I print inside of the wlst script create-wls-domain.py the value of admin_pass it is always blank.

brunoborges:
I also felt like this PR introduces errors instead of fixing something.

@bluezd, mind to please clarify?

bluezd:
@mriccell Without the patch the admin_pass is always blank and the image could not be built successfully, because Docker doesn't allow pass "PASSWORD" so I change the variable name. In my environment the patch definitely fixed this issue even though I specify the --build-arg ADMIN_PASS or not. If I do not specify ADMIN_PASS while building the admin_pass is "welcome1" and if I specify the ADMIN_PASS while building "--build-arg ADMIN_PASS=abcd" the admin_pass is "abcd", below is the debug info:

`sudo docker build -t test --build-arg  .`
```
Initializing WebLogic Scripting Tool (WLST) ...

Welcome to WebLogic Server Administration Scripting Shell

Type help() for help on available commands

### admin_pass = welcome1 ###
domain_name : [base_domain]
```

`sudo docker build -t test --build-arg ADMIN_PASS=abcd .`
```
Initializing WebLogic Scripting Tool (WLST) ...

Welcome to WebLogic Server Administration Scripting Shell

Type help() for help on available commands

### admin_pass = abcd ###
```

```
# docker --version
Docker version 17.03.1-ce, build c6d412e
```

I am curious why it doesn't work in your environment, Could you please re-test it again or provide something more ?

bluezd:
@brunoborges for commit 782ac31 please see my previous comment.
for commit 935a613 it is simple as I explained in the commit, without this patch there is a failure while building the image:
`/bin/sh: /u01/oracle/wlst: Permission denied`

This patch fix this issue.

brunoborges:
Our requirement is that the Dockerfile must not have default passwotd, and a password must always be provided (or generated) during build.

Your change introduces default password.

bluezd:
@brunoborges 
OK, thanks for your info, I will modify the patch to remove the default password. I have two questions as below:

1.  for commit 935a613 if you feel OK with it could you merge it ?
2. for commit 782ac31 I will modify it but I am not clear how to submit it, just post the changed patch here or open another pull request ?

brunoborges:
Create a separate PR for each commits.

PR #435, opened 2017-07-03
12212-domain: Make user setting Weblogic password successfully

bluezd:
While specifying the `--build-arg ADMIN_PASSWORD=welcome1` option for building image the variable `ADMIN_PASSWORD` defined in Dockerfile will never be set due to Docker known mechanism block this behaviour. It causes the image will not be built successfully, so this patch intends to fix this issue. User can specify the password or not. If not the building process will generate the password automatically.

### DIFF
--- a/OracleWebLogic/samples/12212-domain/Dockerfile
+++ b/OracleWebLogic/samples/12212-domain/Dockerfile
@@ -25,7 +25,7 @@ MAINTAINER Monica Riccelli <monica.riccelli@oracle.com>
 
 # WLS Configuration (editable during build time)
 # ------------------------------
-ARG ADMIN_PASSWORD
+ARG ADMIN_PASSD
 ARG DOMAIN_NAME
 ARG ADMIN_PORT
 ARG CLUSTER_NAME
@@ -35,6 +35,7 @@ ARG PRODUCTION_MODE
 # WLS Configuration (editable during runtime)
 # ---------------------------
 ENV ADMIN_HOST="wlsadmin" \
+    ADMIN_PASSWORD="${ADMIN_PASSD}" \
     NM_PORT="5556" \
     MS_PORT="7001" \
     DEBUG_PORT="8453" \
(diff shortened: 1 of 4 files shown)


bluezd:
@mriccell Thanks for your info, I have refined the code as your suggestions, but I have one question in README file:

> On the first startup of the container a random password will be generated for the Administration of the domain. You can find this password in the output line:

The password will be generated in the building process not the domain start, so we are not able to see the password in the domain startup phase.

brunoborges:
Password must be set during `docker run`, not during build, or else all containers based on one single image will have the *same password*. Please rework your scripts.

Djelibeybi:
I'm leaning towards rejecting this PR as we don't accept any PRs that set a password during build. Passwords (if required) must only be set during container startup.

bluezd:
@Djelibeybi  I do not change the mechanism at all, the code before I changing is accepting the password during build, please check the code.

Now I am considering whether fixes this issue together, because setting the password during boot is not acceptable.

Djelibeybi:
The existing mechanism either sets the passowrd on container start using `-e` or randomly generates the password. This is acceptable. Setting the password using `--build-arg` or during build is not acceptable as all containers that are started from this image will then have the same  password.

I'm going to reject this PR as it contravenes our security policies. Please open a new PR if you need to fix setting the password on container start.

bluezd:
@Djelibeybi I follow this [https://github.com/oracle/docker-images/tree/master/OracleWebLogic/samples/12212-domain](https://github.com/oracle/docker-images/tree/master/OracleWebLogic/samples/12212-domain) to build the 12212-domain image, the default mechanism is user could specify the password `--build-arg` while building, my patch does't involve it.

Djelibeybi:
Thanks. I'll flag that sample so that it's fixed.

bluezd:
Can I take volunteer to fix it ?  I saw this issue but not sure whether it is acceptable, so not included in my #435 patch.

Djelibeybi:
Absolutely! 👍 Feel free to submit a new PR that fixes this issue securely, as required by our new rules.

brunoborges:
Those samples are old. We'd highly appreciate someone to check on them and contribute proper setup :-)

bluezd:
Great, I will work on this, but not sure whether somebody else work on it as well ... I am not familiar with github bug tracing process, whether it will be assigned to someone ?

Djelibeybi:
Once you submit the PR, we'll assign it to @mriccell who is the WebLogic product manager for review. It's likely that @brunoborges and I will review it too.

mriccell:
You can look at how it is done in the sample script from OracleWebLogic/dockerfiles/12.2.1.2/container-scripts/createAndStartEmptyDomain.sh. The Dockerfile in sample/12212-domain in the CMD should call a script that generates the Admin Password, creates the Domain, and starts the Admin Server using the WLST scripts in 12212-domain/container-scripts.  Let me know if you need help.

mriccell:
I will submit a pull request to make these changes so that the Admin Password is passed at runtime instead of at build time.

PR #444, opened 2017-07-04
Chooice Oracle Base - Oracle Database

susguzman:
You can choose the Oracle Base directory using a build-time variable when you create the Oracle Database image.

### DIFF
--- a/OracleDatabase/dockerfiles/11.2.0.2/Dockerfile.xe
+++ b/OracleDatabase/dockerfiles/11.2.0.2/Dockerfile.xe
@@ -34,8 +34,9 @@ MAINTAINER Gerald Venzl <gerald.venzl@oracle.com>
 
 # Environment variables required for this build (do NOT change)
 # -------------------------------------------------------------
-ENV ORACLE_BASE=/u01/app/oracle \
-    ORACLE_HOME=/u01/app/oracle/product/11.2.0/xe \
+ARG ORACLE_BASE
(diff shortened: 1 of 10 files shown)


brunoborges:
To be honest, I don't see value in a custom base folder. If programs need to get to the base folder, they should just point to $ORACLE_BASE, but having this as a dynamic environment variable is pointless for a container image, IMO.

gvenzl:
Thank you very much for your contribution, @susguzman.

May I ask what the use case is for setting the ORACLE_BASE directory? In a container based environment it shouldn't matter where the binaries are installed and frankly, the user shouldn't care at all about the inside of the container.

Thx,

susguzman:
Thank you for your comments. The key reason was to change the Oracle Base to /u01/app/oracle, that is the standard. I agree with you, gvenzl, the user shouldn't care at all about the inside of the container'.

I changed the code. I removed the option from the file buildDockerImage.sh, now the oracle base is /u01/app/oracle for all installations and if you are an expert user you can change the oracle base setting the ORACLE_BASE built time variable at image creation time. 

Ty.

gvenzl:
Thanks @susguzman, but I still don't see any benefit of this change. It is discouraged to even log into a container so why would anybody care where the software is installed inside it? Containers are not VMs.

Djelibeybi:
I'm closing this as there is no need to change ORACLE_BASE inside a container.

PR #440, opened 2017-07-04
Create core under oracle/OracleTuxedo

DongHan2016:
1. Updated README.md 
2. Deleted dockerfiles and sampels directories under oracle/OracleTuxedo

PR #448, opened 2017-07-06
Oracle/Tuxedo: Make oracle/tuxedo image running commands successfully

bluezd:
We are not able to run the tuxedo related commands(like
tmshutdown/tmboot) while running oracle/tuxedo container.
This patch fixes this issue. It is necessary because user
needs to to do some testings within the container.

Without this patch oracle/tuxedoshm container do not run successfully
as well.

### DIFF
--- a/OracleTuxedo/dockerfiles/12.1.3/Dockerfile
+++ b/OracleTuxedo/dockerfiles/12.1.3/Dockerfile
@@ -13,7 +13,7 @@ MAINTAINER Todd Little <todd.little@oracle.com>
 
 # Core install doesn't include unzip or gcc, add them
 # Create the installation directory tree and user oracle with a password of oracle
-RUN yum -y install unzip gcc file; yum -y clean all && \
+RUN yum -y install unzip gcc file hostname; yum -y clean all && \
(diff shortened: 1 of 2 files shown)


Djelibeybi:
Why are you modifying the existing Tuxedo `Dockerfile` when you have another pull request to replace it? This modification should be added to your other pull request so that it fixes that, instead of the version that's going to be replaced when that PR is merged.

bluezd:
The another pull request is not mine and I do not think they are the same issue. My commit intends to fix a bug.

Djelibeybi:
There is pull request #440 open that replaces the existing Tuxedo image with a new one. As that PR is in progress and the existing images are going to be replaced, I'm going to close this PR. If the new PR doesn't include these fixes, please speak to the author of that PR to get them included.

bluezd:
#440 has not been merged yet, why not take a look at this first ? It do make sense.

brunoborges:
Hi @bluezd, what we are asking is that you wait for PR #440 to be merged so then you can fix the issue pointed here, or, comment on that PR directly and suggest author to fix this on his own PR.

PR #440 has major enhancements and changes to the OracleTuxedo Docker images set, and we are prioritizing that.

bluezd:
Got it, sure, thanks.

PR #453, opened 2017-07-07
Fix grammar

schedutron:
Docker sounds better with "contains" than "stores" and "open source" isn't a proper noun, hence capitalisation removed.


schedutron:
Haha! It's good to know that programmers still care about grammar :)

PR #467, opened 2017-07-12
Clarify rule around SSH daemon.

### LINKED ISSUE #466: The contribution guide needs to be clearer about SSH
The guide is a little ambiguous about what is not allowed inside a container. It should be clarified that the rule is that an SSH daemon (`sshd`) should not be run.

Djelibeybi:
Resolves #466.


Djelibeybi:
Ping @brunoborges: auto-correct "correction" fixed. :)

Issue #469, opened 2017-07-12
OracleDatabase - not enough swap space on Linux build machine


andersjanmyr:
I have the same issue as https://github.com/oracle/docker-images/issues/230, but on a build machine using Linux, `Linux version 3.10.0-514.10.2.el7.x86_64 (builder@kbuilder.dev.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-11) (GCC) ) #1 SMP Fri Mar 3 00:04:05 UTC 2017`

```
Express Edition requires 2048 MB of swap space. This system has 0 MB
of swap space.  Configure more swap space on the system and retry the 
installation.
error: %pre(oracle-xe-11.2.0-1.0.x86_64) scriptlet failed, exit status 1
error: oracle-xe-11.2.0-1.0.x86_64: install failed
Removing intermediate container 07b9541d2e9e
Running cleanup...
Removing tagged image 'docker.prod.walmart.com/oracle:11.2.0.2-xe'.
Tagged image doesn't exist.
```

Why is the swap space checked at build time? The image may be run on a different machine containing enough swap space.

Djelibeybi:
The swap space is tested by the RPM pre-install script which is triggered by the install of the RPM regardless of platform. We can't modify how the RPM works, we just have to handle it inside the build process which is where the RPM is installed.

andersjanmyr:
@Djelibeybi Okay, thanks. I figured it was something like that. Any other way to trick Docker that it has more swap space available?

Djelibeybi:
I don't believe so. I know you can modify `/dev/shm` but not swap itself. However, let's ping @gvenzl in case he knows.

A workaround/hack that may work is to create a swapfile and turn it on within the build process prior to installing the RPM. I've not tested this, though. :)

gvenzl:
There is several discussions on Docker and swap space like this thread on their [Discussion Forum]( https://forums.docker.com/t/docker-swap-space/3908).

I could never figure out what good way exists to emulate sufficient swap inside the container without artificially blowing up the image size, etc. As of today I'm only aware that on Linux the swap is inherited (I think):

```
[root@localhost oradata]# df -h
Filesystem           Size  Used Avail Use% Mounted on
devtmpfs             3.9G     0  3.9G   0% /dev
tmpfs                3.9G  128K  3.9G   1% /dev/shm
tmpfs                3.9G  8.9M  3.9G   1% /run
[...]

PR #473, opened 2017-07-12
Update OracleDatabase/README.md

brunoborges:
Fix for issue #469 with minimum requirements for building Oracle Database image for Docker.

### DIFF
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -12,6 +12,12 @@ To assist in building the images, you can use the [buildDockerImage.sh](dockerfi
 
 The `buildDockerImage.sh` script is just a utility shell script that performs MD5 checks and is an easy way for beginners to get started. Expert users are welcome to directly call `docker build` with their prefered set of parameters.
 
+### Requirements for Building
+If you intend to build, make sure you have:
+
+1. Enough Docker storage space: 20 GB minimum
+1. Enough Swap memory: 1 GB minimum
+
 ### Building Oracle Database Docker Install Images
 **IMPORTANT:** You will have to provide the installation binaries of Oracle Database and put them into the `dockerfiles/<version>` folder. You only need to provide the binaries for the edition you are going to install. The binaries can be downloaded from the [Oracle Technology Network](http://www.oracle.com/technetwork/database/enterprise-edition/downloads/index.html). You also have to make sure to have internet connectivity for yum. Note that you must not uncompress the binaries. The script will handle that for you and fail if you uncompress them manually!
 
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -13,10 +13,10 @@ To assist in building the images, you can use the [buildDockerImage.sh](dockerfi
 The `buildDockerImage.sh` script is just a utility shell script that performs MD5 checks and is an easy way for beginners to get started. Expert users are welcome to directly call `docker build` with their prefered set of parameters.
 
 ### Requirements for Building
-If you intend to build, make sure you have:
+In order to successfully build this image, please ensure your host environment has:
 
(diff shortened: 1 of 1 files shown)


OracleDatabase/README.md
```diff
@@ -12,6 +12,12 @@ To assist in building the images, you can use the [buildDockerImage.sh](dockerfi
 
 The `buildDockerImage.sh` script is just a utility shell script that performs MD5 checks and is an easy way for beginners to get started. Expert users are welcome to directly call `docker build` with their prefered set of parameters.
 
+### Requirements for Building
+If you intend to build, make sure you have:
+
+1. Enough Docker storage space: 20 GB minimum
```
Djelibeybi: It might be worth actually calling out how to verify this, particularly if the device-mapper storage driver is in use. Or at least link to the documentation.

OracleDatabase/README.md
```diff
@@ -12,6 +12,12 @@ To assist in building the images, you can use the [buildDockerImage.sh](dockerfi
 
 The `buildDockerImage.sh` script is just a utility shell script that performs MD5 checks and is an easy way for beginners to get started. Expert users are welcome to directly call `docker build` with their prefered set of parameters.
 
+### Requirements for Building
+If you intend to build, make sure you have:
```
Djelibeybi: I would rewrite this to: "in order to successfully build this image, please ensure your build host has:"

OracleDatabase/README.md
```diff
@@ -12,6 +12,12 @@ To assist in building the images, you can use the [buildDockerImage.sh](dockerfi
...
+
+1. Enough Docker storage space: 20 GB minimum
+1. Enough Swap memory: 1 GB minimum
```
Djelibeybi: Change memory to space, i.e. "Enough swap space"

gvenzl:
Unfortunately those requirements aren't generally applicable.

The swap space is **only** required for the 11g XE image and it will remain to be the only build for which this remains true.

The storage requirements are also version depending:
* 11g XE needs 400MB
12.1 needs 12GB
12.2 needs 15GB
With the 17.03 release on Oracle Linux we now have the `sqash` experimental available which may drives the requirement even further down but tests are still pending for that.

The only right (and manageable over time) approach to me is to check all the requirements within the Docker build itself. The [space check](https://github.com/oracle/docker-images/blob/master/OracleDatabase/dockerfiles/12.2.0.1/checkSpace.sh) already exists but seems to not catch every case, so we strive to improve to make that solid proof.
The swap space check for 11g XE is not there today but the error message is sufficient enough IMO. I'd rather work on mocking the swap space if possible as it isn't really needed anymore on modern hardware.

brunoborges:
Well it doesn't heart to set minimum requirements to the large values (1GB swap, 15GB storage). It wil surely reduce the amount of issues asking for help

gvenzl:
I highly doubt that. If people cannot even read
```
Error: The container doesn't have enough memory allocated.
A database XE container needs at least 1 GB of shared memory (/dev/shm).
You currently only have 64 MB allocated to the container.
```
from the build output then I don't think it makes them read the ReadMe either; same with the space check.

Furthermore, the swap space is just wrong information, you don't have to have it unless you want XE. And let's not forget that most modern systems don't even have swap anymore.

brunoborges:
How damaging will it be to have generalized minimum requirements?

gvenzl:
If we tell everybody that they now need 1GB swap space which they don't have configured and don't need... Probably quite a bit. Maybe we can address it in a table rather than just bullet points.

Djelibeybi:
I think a table of requirements sounds like a good idea.

Djelibeybi:
I'm going to close this. If @gvenzl feels the docs need to be updated, he can submit a new PR.

Issue #468, opened 2017-07-12
Changing sqlnet.ora parameter ??


AH0HA:
https://community.oracle.com/thread/4062527

changing sqlnet.ora parameter after the docker image is built WITHOUT rebuilding the docker image ?

Thanks for all the help

Djelibeybi:
Just to clarify: are you using an image from Oracle Container Registry or the Docker Store, or are you manually building an image using the `Dockerfile` from this repository?

AH0HA:
manual built.

gvenzl:
This is currently not supported as the file is not being exposed via the volume. However, that will only be a minor change. Stay tuned!

PR #477, opened 2017-07-14
Address issue #468 and allow for Docker build options

gvenzl:
The main changes here are for issue #468.
But with it also comes a new flag for the build script to allow to pass on build options to the Docker daemon. This is useful for the new experimental `--squash` option which reduces images by half.


OracleDatabase/dockerfiles/buildDockerImage.sh
```diff
@@ -103,7 +111,7 @@ elif [ $EXPRESS -eq 1 ] && [ "$VERSION" != "11.2.0.2" ]; then
   exit 1;
 else
   EDITION="xe";
-  DOCKEROPS="--shm-size=1G";
+  DOCKEROPS="--shm-size=1G $DOCKEROPS";
```
Djelibeybi: Can you double-check the logic here? I have this feeling that you're only adding the `$DOCKEROPS` parameter if you're building the XE image.

gvenzl: No, this is done so that I won't loose other options passed on. The options are set between line 88 and 94 of the script.

Djelibeybi: Ah, gotcha. I misread/misunderstood the point of that block. Thanks for correcting me.

gvenzl: No problem at all! :)

PR #482, opened 2017-07-17
Pushed SOA BPM docker images

ykotecha:
Please review SOA, OSB and BPM docker files. Thanks.


OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+#         Oracle Fusion Middleware 12c (12.2.1.2.0) Business Process Management
+#         Quick Start for Developers
+#       - V779125-01_2of2.zip
+#         Oracle Fusion Middleware 12c (12.2.1.2.0) Business Process Management
+#         Quick Start for Developers
+#       - Unzip to get the 2 jars mentioned above
+#
+# Pull base image
+# ---------------
+FROM container-registry.oracle.com/java/serverjre:8u131
```
Djelibeybi: This needs to be changed to `FROM oracle/serverjre:8`.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+#
+# Maintainer
+# ----------
+MAINTAINER Prashanth Nagaraja <prashanth.nagaraja@oracle.com>
+#
+#
+# Environment variables required for this build (do NOT change)
+# -------------------------------------------------------------
+USER root
```
Djelibeybi: This line isn't necessary.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+COPY *.response oraInst.loc /u01/
+RUN cd /u01 && \
+  $JAVA_HOME/bin/java -jar $FMW_JAR1 -silent -responseFile /u01/soabpm.response -invPtrLoc /u01/oraInst.loc -jreLoc $JAVA_HOME -ignoreSysPrereqs -force -novalidation ORACLE_HOME=$ORACLE_HOME && \
+  rm -f /u01/*.jar /u01/oraInst.loc /u01/*.response
+
+#
+# Copy Domain creation scripts
+#
+USER root
+COPY container-scripts/* /u01/oracle/dockertools/
```
Djelibeybi: Move lines 84-85 to after line 72 so that you don't have to switch between `root` and `oracle`.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+RUN chmod a+xr /u01/oracle/dockertools/*.*
+
+#
+# Define default command to start bash.
+#
+USER oracle
```
Djelibeybi: If lines 84-85 are moved, you don't need to switch back to `oracle` here.

OracleSOABPM/dockerfiles/README.md
```diff
@@ -0,0 +1,33 @@
...
+1. **FIRST TIME ONLY**
+   - Update the OracleSOABPM/setenv.sh file to point
+     to the right locations for the following environment variables
+     specific to your environment (See top of file)    
+     DC_USERHOME, DC_REGISTRY_SOA, DC_REGISTRY_DB    
+   - Update the OracleSOABPM/setenv.sh file with the proxy details
+     if you have an internal proxy - ensure that the proxy
+     related env variables are set    
+     http_proxy, https_proxy, no_proxy
+   - Pull the base image into your local docker cache for builds.
```
Djelibeybi: The preferred mechanism for manually building images is to manually build the base image. See the WebLogic Dockerfiles for examples on how to either manually build the `oracle/serverjre` image or pull it from the Oracle Container Registry or Docker Store and tag it accordingly.

OracleSOABPM/dockerfiles/README.md
```diff
@@ -0,0 +1,33 @@
...
+     You will need to provide your login credentials.
+
+    ```sh
+docker pull container-registry.oracle.com/java/serverjre:8u131
```
Djelibeybi: Specifying `8u131` is too specific, because it will lock users into a JRE version that won't be updated. You should specify `8` instead so that when the JRE is updated, any subsequent builds of SOA/BPM will use the updated JRE as well.

OracleSOABPM/setenv.sh
```diff
@@ -0,0 +1,70 @@
...
+  export DC_HOSTNAME=`hostname -f`
+  #
+  # Used by Docker Compose from the env
+  # Oracle DB Parameters
+  #
+  export DC_ORCL_PORT=1521
+  export DC_ORCL_OEM_PORT=5500
+  export DC_ORCL_SID=soadb
+  export DC_ORCL_PDB=soapdb
+  export DC_ORCL_SYSPWD=welcome1
```
Djelibeybi: We don't allow hard-coded passwords anywhere. This needs to be changed to either accept a password or dynamically generate one.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
mriccell: Why are you not extending the FMW Infrastructure image, taking advantage of the layering instead of reinstalling the same binaries here?

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: The reason for the switch was we found that everytime there is a change in one of our scripts an entire reinstall was done as the layer was marked as changed - so the cache was invalidated. This resulted in a really long build time for each minor change in the script. By doing this we get the benefit of caching. The developer will face this too, which could be an issue. The reason for the change was a much better user experience.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: I did not see 8 in container-registry - that is why used this so that we can pull an already existing certified image. Can be changed easily, but local tags would need to be done then.

OracleSOABPM/dockerfiles/README.md
rajlali: I did not see 8 in container-registry - that is why used this so that we can pull an already existing certified image.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: Our Quickstart installers already include FMW infrastructure in them - so they directly do the install. Otherwise it will be a double install.

OracleSOABPM/setenv.sh
rajlali: We support auto password generation - we can remove this.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
Djelibeybi: Do you have alternative installers that don't include FMW infrastructure? If so, can you use those instead?

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: Just as a reference point if we made a change to container-scripts directory which is primarily where all code is, if we did not split it up a build takes around 900 seconds, whereas with this shift it comes down to about 100 seconds.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
Djelibeybi: There is definitely an 8 in the Container Registry, but either way, you need to use `oracle/serverjre:8` for Dockerfiles in this repo.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
Djelibeybi: Once your PR is merged, your scripts wouldn't change, so this is not impacted. Besides, making an efficient image with the least amount of layers is better than jumping between users and layers.

OracleSOABPM/dockerfiles/README.md
Djelibeybi: It's there. :)

OracleSOABPM/samples/README.md
```diff
@@ -0,0 +1,61 @@
...
+    Server
+  - Subsequent runs should simply start the Admin Server
+  - Use **soaas|bpmas|osbas** depending on which one you want to start
+
+    ```sh
+docker-compose up -d soaas
+docker logs -f soaas
+    ```
+  - Verify Admin Server  
+    http://host.acme.com:7001/console (weblogic/welcome1)
```
rajlali: These are just examples for the user. They need to substitute their actual values (of host and username/password etc)

OracleSOABPM/dockerfiles/README.md
```diff
@@ -0,0 +1,33 @@
...
+   - Update the OracleSOABPM/setenv.sh file to point
+     to the right locations for the following environment variables
+     specific to your environment (See top of file)    
+     DC_USERHOME, DC_REGISTRY_SOA, DC_REGISTRY_DB    
+   - Update the OracleSOABPM/setenv.sh file with the proxy details
+     if you have an internal proxy - ensure that the proxy
+     related env variables are set    
+     http_proxy, https_proxy, no_proxy
+   - Pull the base image into your local docker cache for builds.
+     You will need to provide your login credentials.
```
Djelibeybi: This shouldn't be required if the primary documentation states that the `oracle/serverjre:8` is manually built using the Dockerfile from the repo, instead of being pulled from OCR or Docker Store.

rajlali: This is more an internal thing here - though users may face it too. Basically the proxy values not being set results in very slow downloads as it contacts the proxy for internal URLs too. This is again generic stuff that we found along the way as we were doing work - and documented it for others that may run into the same issue.

OracleSOABPM/dockerfiles/README.md
rajlali: Easy change.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: Yes, but that is a lot of separate installers - this was built purely on the request of customers as they wanted to get done in a single download.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: We had exactly this before - and it was just becoming difficult to iterate locally. I assume after merging the developer is likely to change the scripts - so they will face the issue if they do this. It is a simple change so it can be done - but I wanted to make sure you understand the rationale.

Djelibeybi: Why would a developer be likely to change these scripts after merge? That bit I don't understand. These scripts should be production-ready and not require any modification after merging.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
Djelibeybi: Customers are going to build this once and then distribute internally. The images we provide on OCR and other registries will be built off the FMW Infrastructure image. This needs to change to inherit from FMW Infrastructure, because that's exactly what it was designed for: to be the base for SOA/BPM/OSB.

OracleSOABPM/samples/README.md
Djelibeybi: If they're examples, then the documentation needs to be more explicit about that and what the user needs to replace them with.

rajlali: We will just remove it.

OracleSOABPM/dockerfiles/README.md
brunoborges: Please make sure the Dockerfiles and scripts here work for the non-employee user.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: FIXED.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: FIXED.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: FIXED.

OracleSOABPM/dockerfiles/README.md
rajlali: NOTED.

OracleSOABPM/dockerfiles/README.md
rajlali: Yes they will. OracleSOABPM renamed to OracleSOASuite.

OracleSOABPM/samples/README.md
rajlali: FIXED.

OracleSOABPM/setenv.sh
rajlali: FIXED.

OracleSOABPM/setenv.sh
```diff
@@ -0,0 +1,70 @@
...
+  #
+  export DC_ORCL_DBDATA=${DC_USERHOME}/dbdata
+  #
+  # AdminServer Password
+  #
+  export DC_ADMIN_PWD=welcome1
+  #
+  # RCU Common password for all schemas + Prefix Names
+  #
+  export DC_RCU_SCHPWD=welcome1
```
rajlali: FIXED.

OracleSOABPM/dockerfiles/12.2.1.2-soabpm/Dockerfile
rajlali: FIXED. We are now inheriting from the FMW Infra Image.

rajlali: FIXED. We are now inheriting from the FMW Infra Image. We need the official coordinates to use in the FROM tag. Please provide that. Currently using 
FROM middleware/weblogic:12.2.1.2-infrastructure

OracleSOABPM/dockerfiles/README.md
rajlali: FIXED. No longer applicable as we use the FMWInfra Image.

PR #487, opened 2017-07-18
Update server-jre-8u101-linux-x64.tar.gz.download

brunoborges:
Update to Server JRE 8u131


Djelibeybi:
You need rename the file too.

PR #498, opened 2017-07-19
Reduze image size

gvenzl:
This PR saves around 1GB in image size by removing unnecessary components.


OracleDatabase/dockerfiles/12.2.0.1/installDBBinaries.sh
```diff
@@ -45,12 +45,24 @@ fi;
...
+
+# Install Oracle binaries
 cd $INSTALL_DIR       && \
 unzip $INSTALL_FILE_1 && \
 rm $INSTALL_FILE_1    && \
-$INSTALL_DIR/database/runInstaller -silent -force -waitforcompletion -responsefile $INSTALL_DIR/$INSTALL_RSP -ignoresysprereqs -ignoreprereq && \
-rm -rf $INSTALL_DIR/database && \
+$INSTALL_DIR/database/runInstaller -silent -force -waitforcompletion -responsefile $INSTALL_DIR/$INSTALL_RSP -ignoresysprereqs -ignoreprereq
+
+# Remove not needed components
```
Djelibeybi: Would this not be better implemented with conditionals so users can decide via `--build-arg` which of these features is retained/removed?

Also, what is the impact of the removal of these components? I think the documentation should be updated so that you call out what functionality is disabled by each of these removals.

gvenzl: The changes have no impact to the container user::

- $ORACLE_HOME/apex: This is a default APEX installation which is out of date and not enabled in the current container anyways. In future APEX will be provided as a separate Docker image that builds upon this base image, as it should be.
- $ORACLE_HOME/jdbc holds the JDBC client driver jar files and is only needed if somebody wants to run a Java program from the database server, not applicable in a container based environment
 - $ORACLE_HOME/lib/ra*.zip: Those are ZDLRA specific installer files, not applicable for this container
- $ORACLE_HOME/ords: This is a default ORDS installation which is out of date and not enabled in the current container anyways. There is already an ORDS image underway.
- $ORACLE_HOME/sqldeveloper: This is a default SQL Developer installation which is out of date and given that it's a GUI based IDE, not applicable in a container based environment.
- $ORACLE_HOME/ucp holds the UCP client driver jar files and just like JDBC only needed if somebody wants to run a UCP based Java program on the database server, not applicable in a container based environment.

Djelibeybi: That's all I needed to know. :) Is it worthwhile documenting that these components have been removed by default, just in case? I'm going to merge regardless, but something to consider for a follow-up PR.

brunoborges:
LGTM

PR #501, opened 2017-07-20
OracleCoherence: Fix the permission issue while running the container

bluezd:
Got the following error while running the Coherence container:

  "sh: /start.sh: Permission denied"

This patch intends to fix it.

### DIFF
--- a/OracleCoherence/dockerfiles/12.2.1.0.0/Dockerfile.quickinstall
+++ b/OracleCoherence/dockerfiles/12.2.1.0.0/Dockerfile.quickinstall
@@ -49,7 +49,7 @@ COPY extend-cache-config.xml           $COHERENCE_HOME/conf/extend-cache-config.
 
 RUN useradd -b /u01 -m -s /bin/bash oracle && \
    echo oracle:oracle | chpasswd && \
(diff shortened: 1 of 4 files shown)


Djelibeybi:
I'm not sure why there's a difference in using `+x` vs `755`. But, I don't think this is the best solution either and it highlights that these Dockerfiles are not aligned with our best practice around not setting fixed passwords during build.

Let me notify the Coherence maintainer to review all the Dockerfiles and bring them more into alignment. I'm not going to close this PR yet, but I want a larger PR that fixes more issues to be submitted instead.

bluezd:
@Djelibeybi There is no "read" permission of this file. So it can not be executed even though it has "x" permission.

Djelibeybi:
In which case, the permission of the file in git should be fixed too.

bluezd:
@Djelibeybi In general one commit fix one issue, and it is a critical issue because the container could not be started successfully without this patch.

bluezd:
@Djelibeybi I do not think so, I think fix it in Dockerfile is a good way.

Djelibeybi:
Yeah, good point. Merged.

PR #504, opened 2017-07-20
Image creation for Oracle GoldenGate

sbalousek:
This commit contains the infrastructure needed to create Docker images for Oracle GoldenGate.


OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+#
+# HOW TO BUILD THIS IMAGE
+# -----------------------
+#
+# Execute the "buildDockerImage.sh" shell script, with the name of the downloaded OGG filename.
+# Example:
+#      $ ./buildDockerImage.sh ~/Downloads/fbo_ggs_Linux_x64_shiphome.zip
+#
+
+ARG         BASE_IMAGE=container-registry.oracle.com/database/instantclient:12.2.0.1
```
brunoborges: FROM should be the first line in the Dockerfile. Just a matter of standards and code format we are following

OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# Additional installation for Oracle GoldenGate Microservices Architecture
+#
+RUN         [[ "${OGG_EDITION}" != "microservices" ]] && exit 0 || \
+                yum -y --enablerepo *software_collections install openssl rh-nginx18 java-1.8.0-openjdk-headless && \
+                yum -y clean  all && \
+                ln -s /etc/opt/rh/rh-nginx18/nginx /etc/nginx && \
+                sh /etc/ssl/certs/make-dummy-cert  /etc/nginx/ogg.pem && \
+                mkdir -p              ${OGG_DEPLOY_BASE} && \
+                chown oracle:oinstall ${OGG_DEPLOY_BASE}
+
```
brunoborges: Combine multiple RUN calls into one single RUN call.

OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+ADD         ${OGG_TARFILE}         ${OGG_HOME}/
+COPY        runOracleGoldenGate.sh ${OGG_HOME}/
```
brunoborges: Copy artifacts before RUN commands. Ideally, use either one ADD or one COPY.

OracleGoldenGate/Dockerfile
sbalousek: No problem.  I changed the logic to use "if/then/fi" statements which seemed a little clearer.

OracleGoldenGate/Dockerfile
sbalousek: Done and done.  Good catch.

OracleGoldenGate/Dockerfile
sbalousek: Since Oracle GoldenGate is delivered for a wide variety of databases, it's likely that the end-user will want to use a base image other than the default I have here.  For example, Oracle GoldenGate for MySQL doesn't need the Oracle Instant Client base image, which is what I am using for the default.

In order to allow a different base image to be used, a parameter can be used in the FROM statement.  But, in order to use variables in FROM, they *must* be declared first using an ARG statement.  See https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact for additional information.

OracleGoldenGate/README.md
```diff
@@ -0,0 +1,98 @@
...
+ * Oracle GoldenGate 12c Release 2 (12.2.0.1)
+ * Oracle GoldenGate Standard Edition 12c Release 3 (12.3.0.1)
+ * Oracle GoldenGate Microservices Architecture 12c Release 3 (12.3.0.1)
+
+To build the images, use the [buildDockerImage.sh](buildDockerImage.sh) script.
+
+### Building Oracle GoldenGate Docker Images
+**IMPORTANT:** You must download the installation binaries of Oracle GoldenGate. You only need to provide the binaries for the version you plan to install. The binaries can be downloaded from the [Oracle Technology Network](http://www.oracle.com/technetwork/middleware/goldengate/downloads/index.html). You also must have Internet connectivity when building the Docker image for the package manager. Note that you must not uncompress the Oracle GoldenGate binaries. The script will handle that for you and fail if you uncompress them manually!
+
+Once you have downloaded the Oracle GoldenGate software, run the **`buildDockerImage.sh`** script:
```
Djelibeybi: You can't rely on users to use your `buildDockerImage.sh` script. This is provided for convenience but must not be mandatory. Please provide instructions on how to prepare the binaries for a manual build.

This is to ensure that orchestration or CI/CD systems can be used to build the image without needing to run scripts first.

OracleGoldenGate/README.md
```diff
@@ -0,0 +1,98 @@
...
+        -v [<host mount point>:]/u02/ogg \
+        -v [<host mount point>:]/u02/ogg/var/data \
+        oracle/goldengate-microservices:12.3.0.1.0
+
+    Parameters:
+       --name:        The name of the container (default: auto generated)
+       -e OGG_SCHEMA: The GGSCHEMA to use for OGG (default: `oggadmin`)
+       -e OGG_ADMIN:  The name of the administrative account to create for Microservices Architecture (default: `oggadmin`)
+       -e OGG_ADMIN_PWD:
+                      The password for the administrative account (default: value of `OGG_ADMIN`)
```
Djelibeybi: Except I only see this automatic password generation for the `microservices` edition. It should be used for the `standard` edition as well.

OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+                yum -y clean  all; \
+            fi; \
+#
+# Additional installation for Oracle GoldenGate Microservices Architecture
+#
+            if [[ "${OGG_EDITION}" == "microservices" ]]; then \
+                yum -y --enablerepo *software_collections install openssl rh-nginx18 java-1.8.0-openjdk-headless && \
+                yum -y clean  all && \
+                ln -s /etc/opt/rh/rh-nginx18/nginx /etc/nginx && \
+                sh /etc/ssl/certs/make-dummy-cert  /etc/nginx/ogg.pem && \
```
Djelibeybi: I see you're generating a dummy cert here, and you have a bunch of CA/certificate related functions in `runOracleGoldenGate.sh`. What I haven 't seen (and I don't know if this is applicable) is how to provide my own SSL certificates. 

If it's possible to use 3rd-party signed SSL certificates, then I think you should document how to provide the certificates into the container at run time and your `runOracleGoldenGate.sh` needs to be updated to handle that.

OracleGoldenGate/README.md
```diff
@@ -0,0 +1,98 @@
...
+**IMPORTANT:** The result will be a Docker image with the Oracle GoldenGate binaries installed. Created images will follow the naming convention of **oracle/goldengate-&lt;edition&gt;:&lt;version&gt;**, for example:
+
+- `oracle/goldengate-standard:12.2.0.1.1`
+- `oracle/goldengate-standard:12.3.0.1.0`
+- `oracle/goldengate-microservices:12.3.0.1.0`
+
+#### Changing the Base Image
+By default, the base image used when building Oracle GoldenGate Docker images is `container-registry.oracle.com/database/instantclient:12.2.0.1`.  This value can be changed by setting the environment variable `BASE_IMAGE` when executing the `buildDockerImage.sh` script. For example:
+
+    BASE_IMAGE="container-registry.oracle.com/database/enterprise" ./buildDockerImage.sh ~/Downloads/fbo_ggs_Linux_x64_shiphome.zip
```
Djelibeybi: Also, you don't explain _why_ or _when_ a user might want to change the base image.

OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+                echo "-- Missing build argument for Dockerfile. See README.md for usage instructions. --" && \
+                echo "----------------------------------------------------------------------------------" && \
+                exit 1; \
+            fi; \
+#
+# Initial setup
+# - Optional proxy setup for Yum
+# - The oracle:oinstall user must use the numeric identifiers 54321:54321
+#
+            [[ -z ${http_proxy} ]] || echo "proxy=${http_proxy}" >> /etc/yum.conf && \
```
sbalousek: Without this change to yum.conf, subsequent invocations of yum will fail when executing `docker build` behind a corporate firewall. Since the yum application does not use the `http_proxy` environment variable, this is the only way to communicate proxy details.

One alternative would be to keep this change and *remove* the added lines from yum.conf once all yum operations complete.

Djelibeybi: Yum uses the `http_proxy` and `https_proxy` variables just fine. That's how all the other images are built: by passing in those settings as `--build-arg`s but leaving the image pristine. It means that the image is built with no assumption as to the target host proxy requirements.

Note that yum only uses `http_proxy` and `https_proxy`. It ignores `HTTP_PROXY` and `HTTPS_PROXY`, so in this case, case matters. :)

OracleGoldenGate/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+                ln -s /etc/opt/rh/rh-nginx18/nginx /etc/nginx && \
+                sh /etc/ssl/certs/make-dummy-cert  /etc/nginx/ogg.pem && \
+                mkdir -p              ${OGG_DEPLOY_BASE} && \
+                chown oracle:oinstall ${OGG_DEPLOY_BASE}; \
+            fi;
+
+EXPOSE      443
+VOLUME     ["${OGG_DEPLOY_BASE}" ]
+WORKDIR     "${OGG_HOME}"
+ENTRYPOINT ["sh", "-c", "exec ${OGG_HOME}/runOracleGoldenGate.sh"]
```
sbalousek: I did it this way because the value for `${OGG_HOME}` can vary, depending on the Oracle GoldenGate edition that is installed.  The use of "sh -c" is necessary so that variable name expansion occurs for ${OGG_HOME}.  If I simply use `ENTRYPOINT ["${OGG_HOME}/runOracleGoldenGate.sh"]`, I'll get this error:

    docker: Error response from daemon: oci runtime error: container_linux.go:262: starting container process caused "exec: \"${OGG_HOME}/runOracleGoldenGate.sh\": stat ${OGG_HOME}/runOracleGoldenGate.sh: no such file or directory".

The script traps signals correctly with this revised ENTRYPOINT:

    ENTRYPOINT ["sh", "-c", "${OGG_HOME}/runOracleGoldenGate.sh"]

Djelibeybi: How does `$OGG_HOME` vary? It is defined via an `ENV` directive as `/u01/app/ogg`. 

Further, is there a reason why `runOracleGoldenGate.sh` needs to be in `$OGG_HOME`? It could be copied into / and this could be simplified to `ENTRYPOINT ["/runOracleGoldenGate.sh"]`.

OracleGoldenGate/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+    docker logs <container name> | head -3
+    ----------------------------------------------------------------------------------
+    --  Password for administrative user 'oggadmin' is 'qVc3bqNlwijk'
+    ----------------------------------------------------------------------------------
+
+#### Running GGSCI in an OGG Standard Edition Docker container
+The **GGSCI** utility can be run in the OGG container with this command:
+
+    docker exec -ti --user oracle <container name> ggsci
```
sbalousek: The above uses `docker exec`, but I think your comments apply instead to `docker run`.

Oracle GoldenGate comes with a command line utility:

- `ggsci` for Standard Edition
- `adminclient` for Microservices Architecture

The `ggsci` utility *must* be run from within the container because of on-disk state information shared with the background processes and other reasons.

The `adminclient` could be run from another container, but it seemed better to provide the same `docker exec` mechanism for both the Standard and Microservices editions.

Djelibeybi: I was talking about `docker exec`, but I was wrong: it doesn't actually fire the `ENTRYPOINT`, which is curious. Apologies! Ignore this comment.

OracleGoldenGate/Dockerfile
sbalousek: I placed it in ${OGG_HOME} based on @brunoborges comment regarding combining the ADD and COPY into one operation.

For simplicity's sake, I'll place the script in the filesystem root and clean all this up.

OracleGoldenGate/Dockerfile
sbalousek: Huh. So it does. I've used this convention many times - perhaps it's a limitation in older versions of Yum.  Anyway, thanks for this.

OracleGoldenGate/README.md
sbalousek: The Standard Edition does not use passwords.  The only way to use GoldenGate Standard Edition is by running the `ggsci` utility from within a running container.

Djelibeybi: In which case, the default is randomly generated, not the value of `OGG_ADMIN`.

sbalousek: Right.  This documentation is incorrect.

OracleGoldenGate/Dockerfile
sbalousek: Agreed that this area needs better documentation.  My intent was originally to generate the dummy certificate at build time and allow the user to change it at run time by using Docker `--volume`.

A more secure solution would be to generate the dummy certificate at run time and **not** build time.

Djelibeybi: I have no problem with the dummy certificate. Allowing the user to change it at runtime using a `--volume` is exactly what I want. All that's missing is the documentation on how to do this. :)

OracleGoldenGate/Dockerfile
```diff
@@ -71,7 +69,8 @@ RUN         if [[ -z "${OGG_VERSION}" || -z "${OGG_EDITION}" || -z "${OGG_TARFIL
 # Additional installation for Oracle GoldenGate Microservices Architecture
 #
             if [[ "${OGG_EDITION}" == "microservices" ]]; then \
-                yum -y --enablerepo *software_collections install openssl rh-nginx18 java-1.8.0-openjdk-headless && \
+                . /etc/os-release && MAJOR_VERSION="${VERSION/.*/}" && \
```
Djelibeybi: This isn't really necessary: we're not going to switch from `oraclelinux:7-slim` as the base image so we know that the `MAJOR_VERSION` will be `7`. I appreciate the attention to detail, but it's just another level of obfuscation that impedes readability (loquacious pun intended).

brunoborges:
@Djelibeybi, I will take care of merging this once changes are made.

Djelibeybi:
@brunoborges thanks. Remember to "Squash and merge" and combine all the commit messages into a single one. We don't need all the details either during the squash.

Djelibeybi:
@brunoborges and @sbalousek: I've approved this because all the rest of my change requests are cosmetic or best practice, but not blocking. I believe that @sbalousek will fix them anyway, but I didn't want the merge to be delayed in case I'm on a plane and you need my approval to merge.

sbalousek:
@brunoborges: I've committed all the suggested changes except for those in README.md, which I plan to wrap up in the next day or so.

sbalousek:
@brunoborges: I've finished with the suggested changes.

brunoborges:
@sbalousek thanks. Now I just need to get the emails with copyright/license approvals. Send me an email if you need more information.

OracleGoldenGate/Dockerfile
```diff
@@ -68,7 +68,7 @@ RUN         if [[ -z "${OGG_VERSION}" || -z "${OGG_EDITION}" || -z "${OGG_TARFIL
 #    files and must be run from inside the container.
 #
             if [[ "${OGG_EDITION}" == "standard" ]]; then \
-                yum -y install util-linux vi && \
+                yum -y install util-linux vi java-1.8.0-openjdk-headless && \
```
brunoborges: You should not install openjdk from yum. If your Docker image depends on Java, it should extend serverjre/9

sbalousek: Since the Oracle GoldenGate image will likely be created from an OracleInstantClient image, this seemed to be the best solution.

brunoborges: Is Golden Gate certified to run with OpenJDK? If not, you should manually install Oracle Server JRE

sbalousek:
Closed due to lack of corporate approval.

PR #512, opened 2017-07-24
Add a new Dockerfile for Tuxedo Performance Pack

DongHan2016:
1. Updated OracleTuxedo/README.md for the new Dockerfile
2. Added a folder for the new Dockerfile.

### DIFF
--- a/OracleTuxedo/performance_pack/dockerfiles/Dockerfile
+++ b/OracleTuxedo/performance_pack/dockerfiles/Dockerfile
@@ -1,0 +1,16 @@
+# Pull Tuxedo base image
+FROM oracle/tuxedoall:latest
+
+MAINTAINER Judy Liu<judy.liu@oracle.com>
+COPY perfpack_runme.sh /u01/oracle/
+
+USER root
+RUN chown oracle:oracle -R /u01/oracle/ && \
+    chmod +x /u01/oracle/perfpack_runme.sh
+
+USER oracle
+WORKDIR /u01/oracle
+
+# Define ENTRYPOINT. 
+ENTRYPOINT ["/u01/oracle/perfpack_runme.sh"]
+
--- a/OracleTuxedo/performance_pack/dockerfiles/perfpack_runme.sh
+++ b/OracleTuxedo/performance_pack/dockerfiles/perfpack_runme.sh
@@ -1,0 +1,104 @@
+#!/bin/sh
+#
+# Shell script to build and run perfpack.  It assume TUXDIR has been set, or that 
+# Tuxedo has been installed to: ~/tuxHome/tuxedo12.1.3.0.0   If not, invoke 
+# this script with a single argument indicating the location of TUXDIR.
+#
+# Author: Judy Liu
+#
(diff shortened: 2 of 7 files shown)


OracleTuxedo/performance_pack/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Pull Tuxedo base image
+FROM oracle/tuxedoall:latest
```
Djelibeybi: This seems like the wrong `FROM` image. Shouldn't it be `FROM oracle/tuxedo:12.2.1`?

OracleTuxedo/performance_pack/dockerfiles/perfpack_runme.sh
```diff
@@ -0,0 +1,104 @@
...
+# Shell script to build and run perfpack.  It assume TUXDIR has been set, or that 
+# Tuxedo has been installed to: ~/tuxHome/tuxedo12.1.3.0.0   If not, invoke 
+# this script with a single argument indicating the location of TUXDIR.
+#
+# Author: Judy Liu
+#
+# Usage: source perfpack_runme.sh
+#
+cd /u01/oracle/user_projects
+rm -rf perfpack
```
Djelibeybi: If you delete the entire `perfpack` directory on startup, wouldn't this make it impossible to start/stop the container while retaining data? Or is this directory only for ephemeral data?

OracleTuxedo/performance_pack/dockerfiles/perfpack_runme.sh
```diff
@@ -0,0 +1,104 @@
...
+mkdir perfpack
+cd perfpack
+
+# Create environment setup script setenv.sh
+export HOSTNAME=`uname -n`
+export APPDIR=`pwd`
+
+cat <<EOF
```
Djelibeybi: You're outputting this at runtime, but really it should be part of the documentation so that it's set properly in the `docker run` command that actually starts the container.

OracleTuxedo/performance_pack/dockerfiles/perfpack_runme.sh
```diff
@@ -0,0 +1,104 @@
...
+# Boot up the domain
+echo "Boot up the domain..."
+tmboot -y
+# Run the client
+echo "Run the client..."
+./simpcl "If you see this message, perfpack ran OK"
+# Shutdown the domain
+echo "Shutdown the domain..."
+tmshutdown -y
+
```
Djelibeybi: If I'm reading this right, the script that starts at runtime builds a tuxedo client (not sure what that is), boots a domain (don't know what that is), runs the client but doesn't do any error handling in case perfpack doesn't run OK and then shuts down the domain.

I assume at this point the container exits? 

I think the documentation really needs to detail exactly what this container is meant to do and how it does it and how to override the configuration with production-specific information,

DongHan2016:
close it and will pull a new request when get CorpArch approval.

PR #536, opened 2017-08-10
Add UPL 1.0 licensing to all scripts and a COPYRIGHT file at the head…

mriccell:
… directory


Djelibeybi:
@mriccell each commit removes my approval, so let me know when you're done committing and I'll approve then. :)

PR #541, opened 2017-08-11
Adding Oracle REST Data Services Docker image

gvenzl:
Outstanding question still are:

- How to setup networking between the database container and ORDS container in a reliable and recommended manner (@Djelibeybi please advise!)
- New label needed: Oracle REST Data Services


Djelibeybi:
Best practice for Docker networking: https://docs.docker.com/engine/userguide/networking/

OracleRestDataServices/dockerfiles/3.0.10/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+#
+# HOW TO BUILD THIS IMAGE
+# -----------------------
+# Put the downloaded file in the same directory as this Dockerfile
+# Run: 
+#      $ docker build -t oracle/restdataservices:3.0.10 . 
+#
+# Pull base image
+# ---------------
+FROM openjdk:8
```
Djelibeybi: We don't support the official OpenJDK image. Is it possible to use the Server JRE 8 image instead?

OracleRestDataServices/dockerfiles/3.0.10/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+COPY $INSTALL_FILE $CONFIG_PROPS $STANDALONE_PROPS $RUN_FILE $ORDS_HOME/
+
+# Setup filesystem and oracle user
+# Adjust file permissions, go to /opt/oracle as user 'oracle' to proceed with ORDS installation
+# ------------------------------------------------------------
+RUN mkdir -p  $ORDS_HOME/doc_root && \
+    chmod ug+x $ORDS_HOME/*.sh && \
+    groupadd -g 500 dba && \
+    useradd -d /home/oracle -g dba -m -s /bin/bash oracle && \
+    echo oracle:oracle | chpasswd && \
```
Djelibeybi: Setting a hard-coded password is not allowed. This should be removed.

OracleRestDataServices/dockerfiles/3.0.10/Dockerfile
gvenzl: It is possible but what's the issue with openjdk? What implication does it have that it's not supported? We picked it because it allows the user to just pull it from the Docker Hub without having to login anywhere or having to build another image first (similar to what people have today when building on top of the Oracle Linux image.

Djelibeybi: The OpenJDK image is based on Debian Stretch. As such, it has a completely unsupported user space. We cannot provide any support for Oracle customers who may want to run this container if you use the OpenJDK image.

OracleRestDataServices/dockerfiles/3.0.10/Dockerfile
gvenzl: What is the recommended approach for this then? Not setting any password at all? Note, this is also done in the Oracle Database container image.

Djelibeybi: Correct. And I thought we removed it from the DB image. If not, it needs to be removed there as well.

brunoborges:
@gvenzl also, please forward to my email the approval from corparch. We need that in order to merge new project addition.

gvenzl:
This is not a new project, ORDS is part of Oracle Database and also ships with Oracle Database (similar situation as with SQL Developer and APEX). Sure we need corparch approval for that (and APEX and other components coming)?

brunoborges:
Ask CorpArch :-)

brunoborges:
@gvenzl any update?

gvenzl:
Still pending on Corp Arch approval. Meanwhile also the whole PR needs updating for the new ORDS version but luckily that is an easy one.

gvenzl:
@brunoborges, we obtained Corp Arch approval, check your inbox!

gvenzl:
@Djelibeybi, what is the preferred way to add a Unix user within the container? Not set a password at all?

Djelibeybi:
@gvenzl correct.

gvenzl:
Actually, I just double checked and no password is set for the user in the first place.

gvenzl:
Hey @Djelibeybi, now the files are building on top of OracleJava jre 8 out of this repository!

PR #544, opened 2017-08-16
Updated LICENSE and licensing statement.

prkishor:


Djelibeybi:
Did you mean to delete the `startNMandOHS.sh` script?

prkishor:
Yes, I had deleted that in the previous upload only. In this upload, I 
have only done changes related to UPL.

Thanks,

Prabhat

PR #554, opened 2017-08-23
OracleSOASuite: Sync latest oracle/docker-images into personal fork (rajlali/docker-images) 

rajlali:
This pull request attempts to sync the latest oracle/docker-images into my personal fork rajlali/docker-images. So if this pull request lands elsewhere - then feel free to ignore it. I was attempting to do this from the GIThub UI as described in https://stackoverflow.com/questions/20984802/how-can-i-keep-my-fork-in-sync-without-adding-a-separate-remote/21131381#21131381


Djelibeybi:
You don't need to submit a pull request to merge changes from this repo into yours. You just need to set this repo as an upstream remote and then pull with your client.

rajlali:
Avi - Thanks - I was trying to see if there was a way to sync this up using the GUI. But apparently not - which would have been easy and nice - without having to drop down to the command line.

PR #563, opened 2017-09-04
Docker scripts for containing OHS 12.2.1.3.0 installation

prkishor:


mriccell:
Are you removing the 12.2.1.2 image?  Both 12.2.1.2 and 12.2.1.3 images should both be in GitHub.

PR #566, opened 2017-09-06
Create new WLS sample and Correction to the FMWInfrastructure image

mriccell:
Correction to the FMW Infrastructure image to give permissions to all scripts included in the image.
Create new WLS sample extending the WLS image in the DockerStore and connecting to an Oracle DB image also from the Docker Store.
Deploy a WLS Auction application to a single domain Admin Server, deploy a DataSource on the Admin Server that connects to the Oracle database.


brunoborges:
LGTM

PR #568, opened 2017-09-12
managed server listen address is always set to IP

aquasheep20:
When a managed server is created, the listen address is always set to IP address.  This occurs even if the MS_HOST variable is passed.  The behavior can be observed in the WebLogic console.  A problem can arise if the managed server container is restarted and it does not get the same IP.  The result is the managed server will not start.  To fix this the and add-server.py script is updated to set listen address as hostname when the MS_HOST variable is passed.  (the host name should be passed as the container name).  The container IP will be resolved as long as the default docker bridge is not being used.  The existing behavior is unchanged when MS_HOST is not passed,  mshost defaults to msinternal


brunoborges:
@mriccell, mind to take a look at this? If possible test it too?

mriccell:
This change looks good!  @aquasheep20 Thank you so much for contributing!

mriccell:
I looked at the 1213-domain image and in the Dockerfile there is no way for a user to pass in the MS_HOST name.  The MS_HOST should be set in the ENV of the Dockerfile.

aquasheep20:
file removed 
-MZ

brunoborges:
Once @mriccell approves, we merge.

PR #572, opened 2017-09-14
OracleSOASuite: Improve User experience, also setup support for newer release images

rajlali:
- Minor changes to improve user experience
- Add generic support for newer releases


Djelibeybi:
Also, you seem to have changed the attributes of `buildDockerImage.sh` to remove the executable bit. You probably want to put that back, i.e. switch it back to `755` instead of `644`.

rajlali:
Avi - Our Docker code is exactly identical across these releases - the only difference being the Docker file and the checksums. Creating a new directory means replicating content and then maintaining it. I would like to reuse the scripts. Ideas? Please provide the info - and I will review and close this request after that and send a new one based on the discussions.

Thanks. Raj...

Djelibeybi:
How about having a single `Dockerfile` that works with multiple upstream `fmw-infrastructure` versions? You can take a look at how @sbalousek did it with the Oracle GoldenGate images: https://github.com/oracle/docker-images/blob/master/OracleGoldenGate/Dockerfile

Further, you're specifying a `FROM oracle/fmw-infrastructure:12.2.1.3` but that doesn't exist in this repository. This patch set cannot be merged if there isn't a source image to extend yet.

OracleSOASuite/dockerfiles/12.2.1.2/Dockerfile.12.2.1.3
```diff
@@ -0,0 +1,47 @@
...
+# --------------------------
+# This is the Dockerfile for Oracle SOA Suite
+#
+# REQUIRED FILES TO BUILD THIS IMAGE
+# ----------------------------------
+# See *.download file in this directory
+#
+# Pull base image
+# ---------------
+FROM oracle/fmw-infrastructure:12.2.1.3
```
Djelibeybi: This image doesn't exist. The GitHub repo only has a `Dockerfile` for `fmw-infrastructure:12.2.1.2`.

rajlali:
I will close this. I will look at the GG stuff to see if it will work for us - but I would like to optimize. It may not work as our base image version is different - AFAIK it cannot be parameterized. I may just decide to keep the scripts in sync and create 2 dirs if nothing else works.

Closing this request. Will send another one when base image is available. Thanks. Raj...

PR #438, opened 2017-07-03
Container Database 

susguzman:
The new boolean environment variable ORACLE CONTAINER DB allows you choice if you want a multitenant database.

PR #586, opened 2017-09-27
add option for building non cdb database to OracleDatabase.

evantill:


gvenzl:
Thank you very much for your contribution, @evantill.

[The Non-CDB architecture is deprecated since the release of Oracle Database 12c R1 back in 2013.](https://docs.oracle.com/database/121/UPGRD/deprecated.htm#BABDBCJI) Based on that fact we cannot encourage to use, nor want to provide software (tutorials, samples, etc.) that highlight the usage of deprecated functionality. We therefore cannot accept this PR.

Please don't misunderstand this as not being grateful for your contribution or not understanding the use case. We are and we do!

Users are more than welcome to extend the base Oracle Database image and build their own, custom Docker image that does provide a Non-CDB database if that is desired.

As a reference, note that this has already come up once before in PR #438.

evantill:
Understood. Would have be useful for helping people migrating NON CDB database.

PR #590, opened 2017-09-29
New WebLogic on Kubernetes Example

mriccell:
This Example demonstrates the orchestration of a WebLogic 12.2.1.3 domain cluster in a Kubernetes environment. We run through different scaling action of the WebLogic cluster in K8s:
        1- Scale/shrink  the cluster by increasing/decreasing the number of ReplicaSet.
        2- Define a WLDF policy base on Open Session Count MBean and allow the WLS Admin Server initiate the scaling action.


jralmaraz:
Nice! Was waiting for something like this. When can we try it? :)

Djelibeybi:
It was merged a week ago, so you can try it now.

PR #591, opened 2017-10-03
add required files for Oracle Database 11g Release 2 Enterprise Edition

koollman:
basic copy of setup/startup scripts, and simple modification of some values.
Sufficient to get a running database, can probably be much improved


Djelibeybi:
I've added @gvenzl to review but it's unlikely this will be merged as I don't believe we support 11gR2 running inside a Docker container.

I'll wait for his response before reviewing myself.

gvenzl:
@Djelibeybi, you are correct. 11gR2 nor 12.1.0.1 are supported in a Docker environment and therefore won't be made available in this repository.

PR #593, opened 2017-10-04
Fixed a typo in a docker command in README.md

aoakalin:
Changed "Newwork" to "Network", so the line:
   docker inspect --format '{{.NewworkSettings.IPAddress}}' <container-name>
is now:
   docker inspect --format '{{.NetworkSettings.IPAddress}}' <container-name>


Djelibeybi:
Nice catch, thanks.

PR #599, opened 2017-10-11
Fix wrong license statement

brunoborges:
Files are licensed under UPL.


Djelibeybi:
If all the content is under the UPL, do we really need a LICENSE in every folder? Can't we just have one in the root of the repo?

brunoborges:
We could, but for now each project is still evaluated separately.

Djelibeybi:
In which case, only the projects that _don't_ use the UPL need their own `LICENSE` file.

brunoborges:
Let's merge this PR and then remove repeated files. I really want to get done with this :-)

gvenzl:
Fine by me.

PR #617, opened 2017-10-26
Add ORACLE_PDB_COUNT.

ezarko:
Exposes the numberOfPDBs parameter in dbca.rsp.tmpl, including changes to supporting scripts and documentation.

### DIFF
--- a/OracleDatabase/README.md
+++ b/OracleDatabase/README.md
@@ -52,6 +52,7 @@ To run your Oracle Database Docker image use the **docker run** command as follo
 	docker run --name <container name> \
 	-p <host port>:1521 -p <host port>:5500 \
 	-e ORACLE_SID=<your SID> \
+	-e ORACLE_PDB_COUNT=<your PDB name> \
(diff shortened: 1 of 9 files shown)


Djelibeybi:
Ping @gvenzl to review this PR.

gvenzl:
@ezarko, I would like to thank you very much for your contribution! Unfortunately I have to close this pull request though due to following reasons:

1) Creating more than one PDB of a given kind will trigger the Multitenant option. As usual with Oracle software, such choices have to be left to be made by the user explicitly and not hidden behind a parameter that might not further clarify the licensing requirement.

2) Additional PDBs can easily be created by using the setup scripts that will be executed automatically after the database creation. From a runtime point of view it makes no difference whether DBCA creates the PDBs or the scripts afterwards do. However, the scripts allow for better and more fine grained customization such as PDB names, passwords, file locations, etc. See [Running scripts after setup and on startup](https://github.com/oracle/docker-images/tree/master/OracleDatabase#running-scripts-after-setup-and-on-startup) for more details.

PR #638, opened 2017-11-07
Forge /proc and free for successful XE 11.2.0.2 install

piotrowski:
Oracle XE 11.2.0.2 RPM runs into issues on newer Oracle Linux caused by elevated kernel defaults and insufficient container swap. This PR forges these values just for the duration of rpm install, resulting in clean image build.


lastnico:
Worked perfectly when installing it on Docker for Windows 10

Issue #644, opened 2017-11-09
OracleRestDataServices/dockerfiles/buildDockerImage.sh issue with awk on MacOS


martindsouza:
When running `OracleRestDataServices/dockerfiles/buildDockerImage.sh` get the following error:

```bash
awk: syntax error at source line 1
 context is
	match ($0, >>>  /(ords\.)(.{1,2}\..{1,2}\..{1,2})\.(.+.zip)/, <<<
awk: bailing out at source line 1
```

After doing some digging it looks like the 3 parameter `awk` is only available for `gawk`. See [this](https://stackoverflow.com/questions/5466411/print-regexp-matches-in-awk) SO article, specifically the comments on the suggested fix:

> oh, it's gawk... from stackoverflow.com/questions/5536018/… "the awk match() function with three arguments only exists in gawk"

gvenzl:
I'm closing this as the PR has been merged.

PR #645, opened 2017-11-09
#644 fix issue with setting VERSION in macOS bash

martindsouza:
Fix issue (#644) with setting `VERSION` in macOS bash


Djelibeybi:
Ping @gvenzl for approval.

PR #658, opened 2017-11-15
Wls 12213 multihost

trobert57:
Create a Docker Image with a simple WebLogic domain and sample application deployed. Explain the steps to set up services to deploy this image in docker swarm mode. One service is for the AdminServer, another service is for the Managed Servers (set --replicas parameter to the number of managed servers) and a third one for the http server as loadbalancer.


mriccell:
I can not approve this PR.  You need to create a sample which is based with the images and the structure created in this GitHub project.

PR #673, opened 2017-11-28
Removed yum install make from BI 12.2.1.3

sjvp:
Remove make from BI 12.2.1.3 image creation as it silently breaks the installer such that OBI Server will not start.

Added README entry to cover the workaround for installing make prior to manually running the R Installer


OracleBI/dockerfiles/12.2.1.3/Dockerfile
```diff
@@ -50,7 +50,7 @@ COPY $BI_DISTRO_ZIP1 $BI_DISTRO_ZIP2 install.file oraInst.loc /u01/
 USER root
 
 # tools for installing and running BI
-RUN yum -y install unzip hostname make && yum clean all
+RUN yum -y install unzip hostname && rm -rf /var/cache/yum
```
brunoborges: can you move this call to the other RUN statement?

brunoborges: Also, make sure you clean up yum cache after, to keep the image small. See other Dockerfiles for examples

OracleBI/README.md
```diff
@@ -175,7 +175,15 @@ In this mode, the containers are started in parallel and the BI container must w
 4. If the container is restarted, it requires the same set of environment variables passed in again, even though the domain has already been created.
 
 5. Docker health-check is not implemented.
- 
+
+6. Oracle Business Intelligence R Installer will fail as `make` is not installed.
```
brunoborges: If the Dockerfile is designed to avoid this issue, do we need to keep this documented at all?

OracleBI/dockerfiles/12.2.1.3/Dockerfile
Djelibeybi: This cannot be removed because it has to run as `root`. The other `RUN` statement is run as `oracle`.

Djelibeybi: The `rm -rf /var/cache/yum` command removes the `yum` cache.

OracleBI/README.md
brunoborges: Still wondering if this is necessary to be documented.

sjvp: The R Installer != the core product installer.  The Dockerfile needs to correctly run the core product installer, and it will not work properly if make is present without GCC.

The R Installer is an optional manual step that can be used once the container is running.  It adds additional components which cannot be packaged by Oracle due licensing.

It fails if make is not available, hence this comment.

The correct resolution would be for the R Installer to install make - it already installs a bunch of other packages.  However that bug is in the shipped product.  If PM request, I can fix the Dockerfile to install both make and GCC, but I need to cycle that through corparch.

brunoborges: No need to cycle with corparch. Make and GCC are Oracle Linux packages. Just install what is needed for your image to work as best as possible OOTB.

PR #703, opened 2018-01-09
Remove reduntant return code test in OracleDB image build script

### LINKED ISSUE #702: Redundant exit code check in OracleDatabase/dockerfiles/buildDockerImage.sh
The exit code check in OracleDatabase/dockerfiles/buildDockerImage.sh#L171 is redundant.
1. it is prepended by `echo` that will always succeed
2. the command building the docker image will also always succeed due to the OR combination

```sh
docker build --force-rm=true --no-cache=true $DOCKEROPS $PROXY_SETTINGS -t $IMAGE_NAME -f Dockerfile.$EDITION . || {
  echo "There was an error building the image."
  exit 1
}
BUILD_END=$(date '+%s')
BUILD_ELAPSED=`expr $BUILD_END - $BUILD_START`

echo ""

if [ $? -eq 0 ]; then
cat << EOF
  Oracle Database Docker Image for '$EDITION' version $VERSION is ready to be extended: 
    
    --> $IMAGE_NAME
  Build completed in $BUILD_ELAPSED seconds.
  
EOF
```

samekmichal:
Fixes #702


gvenzl:
Thanks for your contribution, @samekmichal. Actually, the error message is a valid one and has to be thrown further up (before the `exit 1`).
Also, I looked at the issue before the PR and just filed the same. :D
There you go, you actually saved me some time and I bluntly didn't see it. :/
However, thank you very, very much for that!

Issue #736, opened 2018-01-30
Container startup fails due to lack of oracle user permissions


luismsousa:
Upon building the 12.1.0.2 ee image and upon the first run, there's an issue whereby the container can't run due to lack of permissions.

Solution: create `dba` group and add the `oracle` user to it.

I'm creating the PR for review.

Docker info:
```DOCKER info:
Containers: 1
 Running: 0
 Paused: 0
 Stopped: 1
Images: 22
Server Version: 18.01.0-ce
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 10
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 89623f28b87a6004d4b785663257362d1658a729
runc version: b2567b37d7b75eb4cf325b77297b140ea686ce8f
init version: 949e6fa
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.4.0-47-generic
Operating System: Ubuntu 16.04.3 LTS
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 3.35GiB
Name: build-machine
ID: JLL2:DZJF:AFB5:J3PZ:54QH:L3AC:UYT5:K4O2:FV5N:WGCP:OQAR:ID3S
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Labels:
 provider=azure
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

WARNING: No swap limit support
```
Docker run command output:
```$ docker run --name db -p 1521:1521 -p 5500:5500 -e ORACLE_SID=ORCLCDB -e ORACLE_PDB=ORCLPDB1 -e ORACLE_PWD=testpwd -e ORACLE_CHARACTERSET=AL32UTF8 -v /home/docker-user/12.1.0.2/dockervolume:/opt/oracle/oradata oracle/database:12.1.0.2-ee
ORACLE PASSWORD FOR SYS, SYSTEM AND PDBADMIN: testpwd

[...]

PR #737, opened 2018-01-30
issue #736 - adding dba group and oracle user to dba group to prevent permissions …

luismsousa:
…problems at startup as described in issue - issue #736


gvenzl:
The `oracle` user and `dba` group are already created by the preinstall rpm with the UID/GID `54321`. Creating the user and group with the default UID/GID of `1000` doesn't solve the underlying issue with host-only mounted volumes. It merely makes the issue go away for users that have a user/group outside the container with `1000` and recreates the issue for folks that have a user/group outside the container with `54321`.

PR #745, opened 2018-02-05
Updating instructions to download the jar and fixing connection string issue.

ashishchalak:
Updating instructions to download the jar and fixing connection string issue.


ashishchalak:
raising a new request

PR #766, opened 2018-02-21
Updated the doc to remove 777 permission

ncleetus:
Incorporated the review comment 'create a new user with the same uid:gid as the oracle user and give restrictive permission of 700'

### DIFF
--- a/OracleSOASuite/dockerfiles/README.md
+++ b/OracleSOASuite/dockerfiles/README.md
@@ -25,7 +25,10 @@ Sample command:
 
 Data volumes are designed to persist data, independent of the container’s lifecycle. The default location of the volume in container is under `/var/lib/docker/volumes`. There is an option to mount a directory from the host into a container as volume. In this project we will use that option for the data volume. The volume will be used to store Database datafiles and WebLogic server domain files.This volume will be created on the host at `/scratch/DockerVolume/SOAVolume/`. Since the volume is created as "root" user, provide read/write/execute permissions to "oracle" user (by providing permissions to "others"), as all operations inside the container happens with "oracle" user login.
 
-    # chmod -R 777 `/scratch/DockerVolume/SOAVolume`
+    #/sbin/useradd -u 1000 -g 1000 oracledocuser
+    #su oracledocuser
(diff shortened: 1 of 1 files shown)


ncleetus:
Avi, please review the changes.

OracleSOASuite/dockerfiles/README.md
```diff
@@ -25,7 +25,10 @@ Sample command:
 
 Data volumes are designed to persist data, independent of the container’s lifecycle. The default location of the volume in container is under `/var/lib/docker/volumes`. There is an option to mount a directory from the host into a container as volume. In this project we will use that option for the data volume. The volume will be used to store Database datafiles and WebLogic server domain files.This volume will be created on the host at `/scratch/DockerVolume/SOAVolume/`. Since the volume is created as "root" user, provide read/write/execute permissions to "oracle" user (by providing permissions to "others"), as all operations inside the container happens with "oracle" user login.
 
-    # chmod -R 777 `/scratch/DockerVolume/SOAVolume`
+    #/sbin/useradd -u 1000 -g 1000 oracledocuser
```
Djelibeybi: This is not the correct sequence of steps because there probably is already a user on your host system with `uid:gid` of 1000. 

To determine if you already have an appropriate user, run:

```
# getent passwd 1000
```

If that returns a username (which is the first field), you should use that user for the following steps. If not, create the `oracle` user manually. I recommend using `oracle` as the username on the host as well (assuming no existing user), so there is reduced confusion.

```
# useradd -u 1000 -g 1000 oracle 
```

Then, you don't need to become that user to create the directory. You can run the following two commands as root:

```
# mkdir -p /scratch/DockerVolume/SOAVolume/
# chown 1000:1000 /scratch/DockerVolume/SOAVolume/
# chmod 700 /scratch/DockerVolume/SOAVolume/
```

ncleetus:
I have modified the doc to include the check to see if the user already exists with uid:gid of 1000

jochen-halwachs:
For me this worked after I have used the default os user (which have per default userid=1000 and must not be oracle) not oracle. Maybe documentation can lead to wrong assumption ?

PR #828, opened 2018-04-06
Overriding httpd.conf

rajesh-g-gupta:
In some situations we need to tune the parameters in httpd.conf. This change will provide for a way to do it.


Djelibeybi:
Sorry, this was merged by mistake and then closed. Please submit this again.

PR #865, opened 2018-05-07
Update contribution guide 

Djelibeybi:
This makes it clear that images must use `oraclelinux:7-slim` when a more specific product image is not applicable. 

The `7-slim` variant has the smallest attack surface because it has the least amount of packages installed and is updated whenever a CVE patch is released for any package within the image.


Djelibeybi:
@gvenzl I'll wait for your sign off on this as well.

PR #875, opened 2018-05-18
Add recommendation about exposing admin port plus minor cleanup in README.md

doxiao:
@mriccell

### DIFF
--- a/OracleWebLogic/samples/12213-webtier-apache/README.md
+++ b/OracleWebLogic/samples/12213-webtier-apache/README.md
@@ -55,7 +55,7 @@ Start an Apache container by calling:
 
 Now you can access the WebLogic Admin Console under **http://localhost/console** (default to port 80) instead of using port 7001. You can access the console from a remote machine using the weblgoic admin server's `<admin-host>` instead of `localhost`.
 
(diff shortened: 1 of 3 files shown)


OracleWebLogic/samples/12213-webtier-apache/README.md
```diff
@@ -82,6 +82,19 @@ To look at the Docker Container logs run:
...
+While it is natural to expose web applications outside a container, exposing administrative features like the WebLogic Administration Console and a T3 channel for WLST should be given more careful consideration. Similar to running a domain in a traditional data center, the same kind of considerations should be taken into account while running a WebLogic domain in a Docker container, for example:
+
+* Only expose those protocols and ports that need to be exposed.
+* Use secure protocols (HTTPS, T3S, and such).
+* Use custom channels to restrict the protocols that are exposed.
+* Is load balancing required?
+* Is certificate-based integrity needed?
+* How will users authenticate?
+* Is the network channel encrypted?
+
```
mriccell: Merge three bellow bullets:

These include various means to controlling access through T3 protocol, such as:

- Running HTTP on a separate port from T3 and other protocols.
- Not exposing T3 ports outside the firewall (i.e., expose only HTTP).
- Not enabling HTTP tunneling for T3.

If it is necessary to expose T3 outside the firewall, using two-way SSL and connection filters to ensure that only known clients can connect to T3 ports.

PR #885, opened 2018-05-23
further edits

rosemarymarano:
Monica, please review the command in step 5 of the MedREc sample: `$ docker build -t 12213-medrec .`
Is it supposed to have the period included in the command? If not, I will remove it.


Djelibeybi:
Yes, that command requires a period. It specifies the current working directory.

rosemarymarano:
OK, will do.

PR #888, opened 2018-05-25
Fix files chown error for weblogic Dockerfile

hoycdanny:


Djelibeybi:
Is there a reported issue for this? If not, please open an issue first so that we can determine whether the usage is correct before applying this fix.

Djelibeybi:
@mriccell can you please review and comment on whether this PR is necessary? Running the `COPY` as the `oracle` user should be sufficient so I'm not sure why this is required.

OracleWebLogic/samples/wls-K8s/wls-12213-domain/Dockerfile
```diff
@@ -45,8 +45,11 @@ ENV MW_HOME="$ORACLE_HOME" \
 
 # Copy scripts
 # --------------------------------
-USER oracle
+USER root
```
mriccell: I agree with Avi, why do you need to be root to copy the files? Are you having ownership problems running the scripts?

Issue #945, opened 2018-07-25
Use Docker multi-stage build to reduce the final image size


elygre:
Since Docker 17.05 it has been possible to use multi stage builds to reduce the size of the final docker image, ref https://docs.docker.com/develop/develop-images/multistage-build/.

The Docker image for Oracle Database 18.3.0-se2, as built by this repository, is 18GB in size. A multi-stage build can reduce this to about half.

gvenzl:
Actually, it can't. Multi-stage builds copy all binaries over as `root` and upon changing the ownership of these files the image will get bloated again.

What does help is the usage of the `--squash` option, which is still in `experimental` status since introduced but works just fine. If the image is built with `--squash` you save about 10GB of the image.

All of this is written up here: https://geraldonit.com/2017/11/13/how-to-create-small-docker-images/

I highly recommend turning on `experimental` and using `--squash`. You can just pass the flag on to `./buildDockerImage.sh -s -o --squash`

gvenzl:
Leaving this open for now until the PR got validated.

sv72:
You can copy the files with predefined owner, so no chown is needed afterwards.
i.e :
COPY --from=build-stage --chown=oracle:oinstall /opt/oracle /opt/oracle

This reduces the image size (for 12.2.0.1-se2 which I'm using now) to 6.3GB.

The --squash option is experimental and is up for discussion to be removed (although this has been going on since 2017 :) : https://github.com/moby/moby/issues/34565

gvenzl:
PR #979 got submitted.

PR #946, opened 2018-07-25
Add Dockerfile for multi-stage build of oracle/database:18.3.0-se2

elygre:
Add a new Dockerfile that uses the multi-stage build feature of
Docker 17.05 (or newer), to avoid having the installation source
file as part of the final image. This reduces the size of the
built image from 18GB to 9GB.

Issue #945 exists for this problem. This merge request implements multi-stage build only for a single docker image (oracle/database:18.3.0-se2), and if successful, should probably also be implemented for (many) other images.

Using the new feature requires using the "-m" option when running buildDockerImage.sh.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/18.3.0/Dockerfile.se2-multistage
+++ b/OracleDatabase/SingleInstance/dockerfiles/18.3.0/Dockerfile.se2-multistage
@@ -1,0 +1,96 @@
+# LICENSE UPL 1.0
+#
+# Copyright (c) 1982-2018 Oracle and/or its affiliates. All rights reserved.
+#
+# ORACLE DOCKERFILES PROJECT
+# --------------------------
+# This is the Dockerfile for Oracle Database 12c Release 2 Enterprise Edition
+# 
+# REQUIRED FILES TO BUILD THIS IMAGE
+# ----------------------------------
+# (1) db_home.zip
+#     Download Oracle Database 18c Standard Edition 2 for Linux x64
+#     from http://www.oracle.com/technetwork/database/enterprise-edition/downloads/index.html
+#
+# HOW TO BUILD THIS IMAGE
+# -----------------------
+# Put all downloaded files in the same directory as this Dockerfile
+# Run: 
+#      $ docker build -t oracle/database:18.3.0-se2 . 
+#
+# Pull base image
+# ---------------
+FROM oraclelinux:7-slim as base
+
+# Maintainer
+# ----------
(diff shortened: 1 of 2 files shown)


Djelibeybi:
@gvenzl can you test this to see if the permissions are correct and that the resulting image actually works (and is smaller)?

gvenzl:
On it, I see already a very interesting option in the `COPY` instruction that wasn't there when introduced: `COPY --chown=oracle:dba `

elygre:
Ah, yes, my version numbering was off. I referenced 17.05, but indeed, 17.09 is needed. 

(Incidentally, there is a proposal for a further improvement to the COPY command on https://github.com/moby/moby/issues/37123, where copying between layers should be able to retain ownership fully, not just being defined. That would solve the possible case of a file tree having multiple owners in multiple places, and would make the Dockerfile more resilient to subtle changes in the runInstaller script)

gvenzl:
That would be indeed an awesome change, and a long waited for one!

gvenzl:
The PR works fine and is verified.

@elygre, this is a great addition to the build files and once again, thanks for making me aware that multi-stage builds have been enhanced to fix these file ownership issue.

I would like to go ahead and replicate this PR against all the Dockerfiles, not just the SE2 18c image. Furthermore, I want to make this the default, as soon as Oracle Linux ships with Docker 18.03 (right now only 17.06), and repurpose the `-m` option to build slim images instead (which remove some files that are not always needed).

The thing is, I really want to accept your PR and contribution to the repo. But, I don't want to ask you to put all that effort in that I'm planning above.

So my question to you is: are you fine if I reject this PR but still am going to implement multi-stage build based on what you have provided? Or, and I fully understand if, do you want to have this PR merged but then I will have to ask you to make changes on top of it and wait for 18.03 on OL?

Btw, I think we have to rerun `yum` again in the new image, otherwise some packages aren't copied over in `/lib64`, etc. Unless you have a good way to get them over as well, maybe just a `COPY /lib, etc` but at that point it becomes cumbersome.

elygre:
I'm happy that this may have been helpful, and don't mind at all if you want to reject this pull request in favor of a more complete solution. Go with the solution that will bring the best results!

Djelibeybi:
FYI, Oracle Container Runtime for Docker 18.03-1 was released last week which includes support for multi-stage builds, including the `--chown` option.

gvenzl:
Closing this PR as the functionality has been provided and merged via superseding PR #979

PR #970, opened 2018-08-26
wls11g

AnushaSandeep:


Djelibeybi:
Not sure what you're trying to achieve, but deleting everything is not the way to do it.

PR #972, opened 2018-08-28
chore: Update README.MD on OracleDatabase

ardasugur:
Just a little typo of versions fixed.


Djelibeybi:
I'm not @gvenzl but I reckon this is OK. Thanks for spotting this.

PR #979, opened 2018-09-13
Leveraging multi-stage builds for Oracle Database Docker images

gvenzl:
This is the PR based on the fantastic job `elygre` did in PR #946.

A couple of things that have been introduced with this:

* Going forward, there will be a minimum Docker version requirement, currently 17.09. The build script `buildDockerImage.sh` will perform that check and abort if the Docker version is below that. 17.09 was the first version that introduced the `--chown` option for the `COPY` command making these multi-stage builds possible.
* For 12.2.0.1 and 18.3.0 the `Dockerfile.ee` and `Dockerfile.se2` have been consolidated into `Dockerfile`
* Database edition is now passed on as a build argument for 12.2.0.1 and 18.3.0
* Bugfix for checking the return value of the `md5sum` checksum command.
* The `-m` option as originally introduced in #946 has not been take over. Multi-stage builds are the default going forward.


OracleDatabase/SingleInstance/dockerfiles/12.1.0.2/Dockerfile.ee
```diff
@@ -47,29 +48,59 @@ ENV ORACLE_BASE=/opt/oracle \
...
+# Start new stage for installing the database
+# -------------------------------------------
+#############################################
+
+FROM base AS builder
+
+ARG DB_EDITION
+
+# Install unzip for unzip operation
+RUN yum -y install unzip
```
Djelibeybi: I was going to comment that you need to add `rm -rf /var/cache/yum` but I guess that's not necessary because you only copy the Oracle binaries out to the final image, correct?

PR #1008, opened 2018-10-09
Oracle GoldenGate on Docker

sbalousek:
- Use if/then statements instead of simple tests for improved
   compatibility with older versions of Bash.

### DIFF
--- a/OracleGoldenGate/runOracleGoldenGate.sh
+++ b/OracleGoldenGate/runOracleGoldenGate.sh
@@ -38,10 +38,13 @@ OGGProcesses="(adminclient|adminsrvr|distsrvr|extract|ggsci|pmsrvr|recvsrvr|repl
 ##
 function generatePassword {
     local password="$(openssl rand -base64 9)-$(openssl rand -base64 3)"
-    [[ "${password}"  == "${password/[A-Z]/_}" ]] && { generatePassword; return }
-    [[ "${password}"  == "${password/[a-z]/_}" ]] && { generatePassword; return }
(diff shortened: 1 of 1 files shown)


Djelibeybi:
Why is this necessary? The `runGoldenGate.sh` script runs inside the container, so the version of bash is known.

sbalousek:
The version of bash is actually not known.  The base image is variable and could be OL6 or OL7.

Djelibeybi:
This discussion is perhaps better done outside of a pull request, but I guess it's vaguely contextual.

Currently the default is to extend the existing `oracle/instantclient:12.2.0.1` image which extends the `oraclelinux:7-slim` image. 

What purpose does providing the ability to alter that base image serve beyond the potential for the final image to fail if the underlying dependencies are not met and thus increasing your support burden?

sbalousek:
Actually, allowing the base image to be changed by the end user *reduces* the support requirement. Regardless, the base image could be `oracle/instantclient:12.2.0.1`, or an image containing the Oracle RDBMS, or even an image crafted for MySQL.

Is there something technically wrong with the requested change?

Djelibeybi:
Nothing is wrong with the change, I was just curious to know why it was needed in the first place.

sbalousek:
Hi @Djelibeybi,

Sorry, but I do not have the permissions to merge these changes. Is that something you can do?

Thanks,
- Steve

Djelibeybi:
Sorry, I thought I merged them when I approved it earlier. My bad.

PR #1009, opened 2018-10-10
Adding GraalVM CE 1.0.0-rc7

mzachh:
Files to build the official GraalVM CE 1.0.0-rc7 docker image.


GraalVM/CE/1.0.0-rc7/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Note: If you are behind a web proxy, set the build variables for the build:
+#       E.g.:  docker build --build-arg "https_proxy=..." --build-arg "http_proxy=..." --build-arg "no_proxy=..." ...
+
+ARG GRAAL_VERSION=1.0.0-rc7
+ENV LANG=en_US.UTF-8
+
+ENV GRAALVM_PKG=https://github.com/oracle/graal/releases/download/vm-$GRAAL_VERSION/graalvm-ce-$GRAAL_VERSION-linux-amd64.tar.gz \
+    JAVA_HOME=/opt/graalvm-ce-$GRAAL_VERSION/ \
+    PATH=/opt/graalvm-ce-$GRAAL_VERSION/bin:$PATH:/opt/rh/llvm-toolset-7/root/usr/bin \
```
Djelibeybi: Please remove the `$JAVA_HOME/bin` from `$PATH` as this is taken care of by the `alternatives` block below.

GraalVM/CE/1.0.0-rc7/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    && yum-config-manager --enable ol7_software_collections \
+    && yum install -y bzip2-devel ed gcc gcc-c++ gcc-gfortran gzip less libcurl-devel make openssl-devel readline-devel tar vi which xz-devel zlib-devel \
+    && yum install -y libcxx libcxx-devel llvm-toolset-7 \
+    && rm -rf /var/cache/yum
+
+RUN set -x \
+    && curl --fail --silent --location --retry 3 ${GRAALVM_PKG} \
+    | gunzip | tar x -C /opt/
+
+RUN alternatives --install /usr/bin/java  java  $JAVA_HOME/bin/java  20000 \
```
Djelibeybi: Please replace this with a loop that creates links for all binaries instead, e.g.

```
for bin in "$JAVA_HOME/bin/"*; do \
		base="$(basename "$bin")"; \
		[ ! -e "/usr/bin/$base" ]; \
		alternatives --install "/usr/bin/$base" "$base" "$bin" 20000; \
done;
```

See https://github.com/docker-library/openjdk/blob/master/11/jdk/oracle/Dockerfile#L37-L41 for context.

GraalVM/CE/1.0.0-rc7/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    && alternatives --install /usr/bin/javac javac $JAVA_HOME/bin/javac 20000 \
+    && alternatives --install /usr/bin/jar   jar   $JAVA_HOME/bin/jar   20000
+
+CMD java -version
```
Djelibeybi: Would it be better to drop users into `jshell` like the OpenJDK does?

GraalVM/CE/1.0.0-rc7/README.md
```diff
@@ -0,0 +1,63 @@
+
+# Supported tags and respective Dockerfile links
+* 1.0.0-rc7 (1.0.0-rc7/Dockerfile)
```
Djelibeybi: I don't think this needs to be documented here as you're already in the `1.0.0-rc7` directory.

GraalVM/CE/1.0.0-rc7/Dockerfile
mzachh: jdk8 has no `jshell`.

GraalVM/CE/1.0.0-rc7/README.md
mzachh: The idea is to add links to other tags, like on https://hub.docker.com/_/openjdk/.

But as this is the first tag, we can remove it for now.

Djelibeybi: Yeah, once you have multiple tags then this makes sense.

MartinBalin:
Question how this is supposed to work when user installs RC1 and adds e.g. Python. Then pyhton is linked into alternatives with RC1. Then RC2 is added to system and user adds Ruby. Ruby is linked to alternatives. Normally I forgot about older RCs. In this case user can get surprising errors when running python and ruby as these are from different GraalVM RCs. In this scenarios putting /bin folder to PATH seems to be giving user what he expects.

Djelibeybi:
What you're describing is an anti-pattern for container use: a user should not be installing additional binaries into an existing container. The correct pattern would be to create a new image that extends  `graalvm-ce:1.0.0-rc7` image and installs the desired binaries during its build. The build process should create the appropriate `alternatives` for the newly installed binaries (or not: in the case of Python, you really don't want to override the system version).

Djelibeybi:
This isn't the right place to discuss it, now that this PR has been merged. You could open an issue if you think using `alternatives` is not the right solution for Graal.

PR #1027, opened 2018-10-27
Correct the Instant Client 18.3 build syntax

cjbj:
Add the missing 'current directory' argument to `docker build`.


Djelibeybi:
Thanks. I meant to do this but decided to spend the weekend in Yosemite National Park instead. 😁

PR #1044, opened 2018-11-06
Removing old demo of WLS on K8S, the prometheus and Grafana samples h…

mriccell:
…ave been moved under the WebLogic Monitoring Exporter project


Djelibeybi:
Can you please update a README to point to the WebLogic Monitoring Exporter project?

mriccell:
Which README?

Djelibeybi:
Whichever one you think is the most appropriate to guide users to where the content is. :)

Issue #1037, opened 2018-11-02
Missing rights to execute install script for 12.2.0.1-EE


lookyflashbecki:
Hi, 
There was a problem with executing install script during build of docker image 12.2.0.1-EE.
It seems that there is missing execution bit set on  installDBBinaries.sh.
Line related to that issue:
83 RUN $INSTALL_DIR/$INSTALL_DB_BINARIES_FILE $DB_EDITION

Fix to have a complete and successful build is to add line as below before switching to user oracle:
**RUN chmod ug+x $INSTALL_DIR/$INSTALL_DB_BINARIES_FILE**

Working part of code:
`# Copy DB install file`
COPY --chown=oracle:dba $INSTALL_FILE_1 $INSTALL_RSP $INSTALL_DB_BINARIES_FILE $INSTALL_DIR/

RUN chmod ug+x $INSTALL_DIR/$INSTALL_DB_BINARIES_FILE

`# Install DB software binaries`
USER oracle
RUN $INSTALL_DIR/$INSTALL_DB_BINARIES_FILE $DB_EDITION`

Regards

gvenzl:
Analysis: The `chmod` was lost with the introduction of multi-stage builds which happens in the stage before the DB binaries installation.

Issue #232, opened 2016-12-12
Oracle Database - Text file busy while installing oracle linux


frenchc:
Just tried to create an oracle DB image on debian jessy. It failed with 

/bin/sh: /opt/oracle/install/setupLinuxEnv.sh: /bin/bash: bad interpreter: Text file busy

Add sync to chmod fixed the issue.

gvenzl:
I'm not sure whether `sync` is a fix to the issue or just a workaround. The error above would also be generated when the file is open for write when you try to execute it.

Is this problem consistently reproducible on your environment?

frenchc:
Consistent problem. Fixed with sync. Doesn't work without.

See here for related discussion https://github.com/docker/docker/issues/9547

msimion:
same os, same problem; fixed with sync; thanks!

gvenzl:
Thanks, added the synch in both build files. PR coming soon.

PR #1046, opened 2018-11-07
Include chmod before calling DB binaries install

gvenzl:
Fix for #1037 


gvenzl:
The moby bug stated in https://github.com/oracle/docker-images/issues/232 is still open. Given I could never reproduce it on my end, I'm hesitant to remove it.

gvenzl:
It did work for some users, otherwise they wouldn't have run into it.
Why make it harder for them again if the solution is so simple? :)

PR #1084, opened 2018-12-13
Update WLS patch for k8s to have correct patches

markxnelson:
Removed patch p27117282_122130_Generic.zip which is not required
Added patch p28298734_122130_Generic.zip which is required (it is a prereq for p28076014_122130_Generic.zip and it supersedes p27117282_122130_Generic.zip)


mriccell:
Mark can you edit the README as well?  I extend from the developer image because it is a smaller image.

PR #1088, opened 2018-12-17
change to work in relation to executing script in able to include in …

CarolynRountree:
Change the path in the build shell scripts to work in relation to from where the script is executing. This allows the build to be run without changing to the specific directory and easily included in other scripts.
Rename the tag on wdt to align with wlst domain home. Build the archive and perform a curl to get the weblogic-deploy-tooling zip from the Dockerfile during docker build.


OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
```diff
@@ -17,7 +17,7 @@
 #
 # Pull base image
 # ---------------
-FROM oracle/weblogic:12.2.1.3-developer
+FROM wlsldi-v2.docker.oraclecorp.com/weblogic:19.1.0.0
```
doxiao: We cannot use 19c image because it has not been released.

mriccell: Carolyne 19.1 is not GA until next year you have to extend the 12.2.1.3 image

Djelibeybi: Also, this is leaking internal Oracle hostnames and versioning information, which is confidential information. In future, please take a little more care before submitted a PR publicly that contains this sort of data.

PR #1112, opened 2019-01-10
Modified Readme for patching sample and domain-creation script

Prabhat-Kishore:


OracleHTTPServer/dockerfiles/12.2.1.3.0/container-scripts/create-sa-ohs-domain.py
```diff
@@ -27,7 +27,6 @@
 # Create OHS System Component by the name ohs1, Configure OHS Listen Port and SSL Port
 # ======================================================================
 cd('/')
-create(ohs_comp_name, 'SystemComponent')
```
mriccell: Could you explain this change?

Prabhat-Kishore: Having this statement was leading to an error during domain creation.
This is because an OHS instance is itself created when creating a standalone domain.

PR #1115, opened 2019-01-12
Changes to Copyright dates, Clarify patches to apply

mriccell:
1-Change copyright dates
2-Clarify difference between Dockerfiles
3- Fix formatting
4- Clarify in Dockerfile to comment out commands when patch 27117282 is not required
5- Fix issue #1116

### DIFF
--- a/OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
+++ b/OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
@@ -17,7 +17,7 @@
 #
 # Pull base image
 # ---------------
-FROM oracle/weblogic:12213-developer
+FROM oracle/weblogic:12.2.1.3-developer
 
 # Maintainer
 # ----------
--- a/OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
+++ b/OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
@@ -17,7 +17,7 @@
 #
 # Pull base image
 # ---------------
-FROM oracle/weblogic:12.2.1.3-developer
+FROM oracle/weblogic:12213-dev-patch-wls-for-k8s
 
 # Maintainer
 # ----------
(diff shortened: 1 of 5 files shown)


OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
```diff
@@ -17,7 +17,7 @@
 #
 # Pull base image
 # ---------------
-FROM oracle/weblogic:12.2.1.3-developer
+FROM oracle/weblogic:12213-dev-patch-wls-for-k8s
```
doxiao: Do we require that the customers who use the docker image alone (instead of being managed by the Operator) also have to use the patched WebLogic 12.2.1.3 image?

OracleWebLogic/samples/12213-patch-wls-for-k8s/Dockerfile.patch-ontop-12213
```diff
@@ -34,18 +34,43 @@ FROM oracle/weblogic:12.2.1.3-developer
...
+# NOTE: That patch (p27117282) is needed only if the WebLogic binary image
+# is created manually from this Github project under `dockerfiles/12.2.1.3`.
+# Comment out the ENV command if not needed.
 ENV PATCH_PKG1="p27117282_122130_Generic.zip"
 ENV PATCH_PKG2="p29135930_122130_Generic.zip"
 
-# Copy supplemental package and scripts
+# PATCH_PKG1 (27117282) is only needed if the WebLogic binary image
+# is created manually from this Github project under dockerfiles/12.2.1.3.
+# Uncomment the command to `Copy Patch 29135930` and comment the command
```
doxiao: Shall we say "Otherwise, uncomment the command....", or " In other use cases, uncomment the command"?

OracleWebLogic/samples/12213-patch-wls-for-k8s/Dockerfile.patch-ontop-12213
```diff
@@ -34,18 +34,43 @@ FROM oracle/weblogic:12.2.1.3-developer
...
+# COPY $PATCH_PKG2 /u01/
+# --------------------------------
+# Copy patches 27117282 & 29135930 
 # --------------------------------
 COPY $PATCH_PKG1 $PATCH_PKG2 /u01/
 
-# Installation of Supplemental Quick Installer
+# PATCH_PKG1 (27117282) is only needed if the WebLogic binary image
+# is created manually from this Github project under dockerfiles/12.2.1.3.
+# Uncomment the statements to `Apply Patch 29135930` and comment the commands
```
doxiao: See the comment above.

OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
mriccell: This is sample is specific to patching for WebLogic on Kubernetes and the image needs to be patched no matter if it is built manually or pulled from Docker Store.  When the patched image is available in Docker Store then it will only be required when the image is built manually.

PR #1120, opened 2019-01-15
Add property to be able to name the image with different names

mriccell:


OracleWebLogic/samples/12213-domain-home-in-image/build.sh
```diff
@@ -4,9 +4,42 @@
...
+# TAG_NAME              - Tag the docker image with this name. This overrides the default of 
+#                         12213-domain-home-in-imag:latest. 
+#
+#                         There are four ways to tag the domain home image using this build script.
+#
+#                           . Do nothing and the image will be tagged with the default name. 
+#                           . Add an IMAGE_TAG variable to the variable file and allow the
+#                             setEnv.sh to manage the tag. Overrides the default tag.
+#                           . Set the TAG_NAME environment variable. Overrides the default tag.
+#                           . Set the ADDITIONAL_BUILD_ARGS to include a tag argument
```
doxiao: I don't think we have implemented the last option: ADDITIONAL_BUILD_ARGS.

OracleWebLogic/samples/12213-domain-home-in-image/build.sh
```diff
@@ -4,9 +4,40 @@
 #
 #Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 #
-script="${BASH_SOURCE[0]}"
-scriptDir="$( cd "$( dirname "${script}" )" && pwd )"
+# TAG_NAME              - Tag the docker image with this name. This overrides the default of 
+#                         12213-domain-home-in-imag:latest. 
+#
+#                         There are four ways to tag the domain home image using this build script.
```
doxiao: change "four ways" -> "three ways".

PR #1132, opened 2019-01-30
WIP: updates to support FMW Infrastructure running in Kubernetes

markxnelson:


Djelibeybi:
Please update this PR with a proper description of what you're changing and why, so the reviewers know what the expected/intended outcome should be.

OracleFMWInfrastructure/README.md
```diff
@@ -128,7 +128,7 @@ The database is created with the default password `Oradoc_db1`. To change the da
```
mriccell: Add description about the new image Dockerfile-domain-in-image.
We build these images with script buildDockerImge.sh you need to change it so that it can build that image as well.

Issue #1138, opened 2019-02-01
WebLogic No X11 DISPLAY variable was set.


itzamnamx:
Running the following command to build the docker container I got the error No X11 DISPLAY variable was set.

sh buildDockerImage.sh -v 12.2.1.3 -d -s

~/git/docker-images/OracleWebLogic/dockerfiles$ sh buildDockerImage.sh -v 12.2.1.3 -d -s
Skipped MD5 checksum.
=====================
Building image 'oracle/weblogic:12.2.1.3-developer' ...
Sending build context to Docker daemon    839MB
Step 1/14 : FROM oracle/serverjre:8
 ---> f4512166cdb7
Step 2/14 : MAINTAINER Monica Riccelli <monica.riccelli@oracle.com>
 ---> Running in 93f2eafc3146
Removing intermediate container 93f2eafc3146
 ---> c37a727db45d
Step 3/14 : ENV ORACLE_HOME=/u01/oracle     USER_MEM_ARGS="-Djava.security.egd=file:/dev/./urandom"     SCRIPT_FILE=/u01/oracle/createAndStartEmptyDomain.sh     PATH=$PATH:/usr/java/default/bin:/u01/oracle/oracle_common/common/bin:/u01/oracle/wlserver/common/bin
 ---> Running in bc4866d83118
Removing intermediate container bc4866d83118
 ---> 5e13ad694c68
Step 4/14 : RUN mkdir -p /u01 &&     chmod a+xr /u01 &&     useradd -b /u01 -d /u01/oracle -m -s /bin/bash oracle
 ---> Running in ea56e3e576a8
Removing intermediate container ea56e3e576a8
 ---> cb5f67cb2025
Step 5/14 : COPY container-scripts/createAndStartEmptyDomain.sh container-scripts/create-wls-domain.py /u01/oracle/
 ---> b55d6c1b37e0
Step 6/14 : ENV DOMAIN_NAME="${DOMAIN_NAME:-base_domain}"     ADMIN_LISTEN_PORT="${ADMIN_LISTEN_PORT:-7001}"      ADMIN_NAME="${ADMIN_NAME:-AdminServer}"     DEBUG_FLAG=true     PRODUCTION_MODE=dev     ADMINISTRATION_PORT_ENABLED="${ADMINISTRATION_PORT_ENABLED:-true}"     ADMINISTRATION_PORT="${ADMINISTRATION_PORT:-9002}"
 ---> Running in a6a6520a03a0
[...]

PR #1160, opened 2019-02-18
Fixt to issue 1138

itzamnamx:
Hi,

Some variables are not set for the silent installation causing the DISPLAY error, I have fixed this and works with Weblogic 12.2.1.3
 
https://github.com/oracle/docker-images/issues/1138

Regards


mriccell:
@itzamnamx can you make change to the installer version?

itzamnamx:
I will do the change and let you know.

Victor

mriccell:
Made changes to fix Display issue in developer installer image in PR #1228.  Thanks so much for your contribution!

PR #1172, opened 2019-02-28
[OracleRestDataServices] Use custom listen port if specified by PORT env variable

fabiocaseri:
Ability to specify a customizable http listen port using PORT env variable (default remains 8888)

Signed-off-by: Fabio Caseri


gvenzl:
Thanks a lot @fabiocaseri for the PR. Why do you think it's necessary to change the port inside the container when the mapping can be done to any port anyway?

fabiocaseri:
Agree, not so necessary, just to give freedom of choice.

gvenzl:
You still got the freedom by just saying `docker run ... -p 5432:8888` or `docker run ... -p 1234:8888`. Being able to specify the port inside the container does not provide any additional flexibility nor value. It just increases the complexity as people could pull an image that uses a random port inside the container

fabiocaseri:
Ok, we can close this one.
Thanks!

PR #1186, opened 2019-03-10
Oracle GoldenGate on Docker

sbalousek:
- Installation of java-1.8.0-openjdk-headless now requires the
   ol7_optional_latest YUM repository


Djelibeybi:
Can you switch this to use `yum-config-manager` instead? That way the repo will remain enabled for any subsequent yum calls.

sbalousek:
I enabled ol7_optional_latest , but not ol7_software_collections which is only needed for the NGinx and scl packages.

PR #1202, opened 2019-03-26
1201 fixes

rishabh20oracle:
The following things have been done as part of this issue:

1. Modified buildDockerImage.sh to accept a new parameter for goldimage name. This provides the user to provide a custom goldimage (which is useful for internal teams at Oracle) instead of being restricted to the one present on OTN.

2. Created a folder goldimage_based_docker which has all files copied from folder 18.3.0 under dockerfiles.

3. Modified the Dockerfile in this newly created folder to accept 2 arguments: major base release version and goldimage name.

4. Modified buildDockerImage.sh to change directory to 'goldimage_based_docker' for versions starting from 18c.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
@@ -12,9 +12,9 @@
 usage() {
   cat << EOF
 
-Usage: buildDockerImage.sh -v [version] [-e | -s | -x] [-i] [-o] [Docker build option]
+Usage: buildDockerImage.sh -v [version] [-e | -s | -x] [-i] -n [gi_name] [-o] [Docker build option]
 Builds a Docker Image for Oracle Database.
-  
+
 Parameters:
    -v: version to build
        Choose one of: $(for i in $(ls -d */); do echo -n "${i%%/}  "; done)
@@ -22,6 +22,7 @@ Parameters:
    -s: creates image based on 'Standard Edition 2'
    -x: creates image based on 'Express Edition'
    -i: ignores the MD5 checksums
+   -n: goldimage_name (for 18c and higher releases)
    -o: passes on Docker build option
 
 * select one edition only: -e, -s, or -x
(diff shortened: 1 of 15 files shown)


Djelibeybi:
Why are you duplicating files? Why can't the gold image functionality be added to the default image so that we're not duplicating ongoing effort? By copying files it means that any change in one has to be done manually in the other and that is not sustainable over time.

Djelibeybi:
Also, if you are an Oracle employee please follow instructions on how to get your GitHub account added to the Oracle GitHub Organisation. If you don't know where to find those instructions, please look me up on ARIA and email me directly.

rishabh20oracle:
> Why are you duplicating files? Why can't the gold image functionality be added to the default image so that we're not duplicating ongoing effort? By copying files it means that any change in one has to be done manually in the other and that is not sustainable over time.

The intention is to remove the version specific folders going forward. I agree that the goldimage functionality can be added to the default image (Dockerfile) itself and that is specifically what I have done, only in a new folder. The older approach was changing the directory to the <version> folder in accordance to the version given by user, while my changes change the directory to the same folder every time (folder name: goldimage_based_docker) irrespective of the version supplied by the user. Once my changes are expected, we can decide to remove the folders that I have copied from.

Djelibeybi:
If this change is important enough, it should be done on the existing files. Don't create technical debt for someone else at some unspecified time in the future to do more work.

rishabh20oracle:
> If this change is important enough, it should be done on the existing files. Don't create technical debt for someone else at some unspecified time in the future to do more work.

I myself will do the necessary changes once discussed/finalised by reviewers here. This pull request was intended for getting my changes reviewed by Gerald and other reviewers and get their comments.

Djelibeybi:
Pull requests exist for code review on the _actual_ code. I would strongly encourage you to close this PR and resubmit with the full set of changes you're proposing so we can review the entire thing and build it ourselves without duplicating effort.

latakannan:
The intent here was to just make it generic so that we do not have to have separate directories for each of the newer versions that come up.  And yes, completely agree, that we should not incur technical debt here and so we would refactor the code to just do this from 19 onwards. We do not have any 19 directories created as yet and so hopefully this would not cause any issues coz there is no prior footprint of this version so far. 
Is this approach ok?

Djelibeybi:
That's up to @gvenzl. I don't mind either way, so it's up to him.

Djelibeybi:

PR #1212, opened 2019-04-04
Adding GraalVM-CE rc15

mzachh:
Preparation for next release.


mzachh:
We still need to discuss how many older RC releases, we want to keep. But I removed now a few.

PR #1218, opened 2019-04-07
Xe Dockerfile cleanup

gvenzl:


OracleDatabase/SingleInstance/dockerfiles/18.4.0/Dockerfile.xe
```diff
@@ -45,26 +45,24 @@ ENV PATH=$ORACLE_HOME/bin:$PATH
...
 # Copy binaries
 # -------------
-COPY $INSTALL_FILE_1 $RUN_FILE $PWD_FILE $CHECK_DB_FILE $INSTALL_DIR/
-COPY $CONF_FILE $INSTALL_DIR/
+COPY $INSTALL_FILE_1 $RUN_FILE $PWD_FILE $CHECK_DB_FILE $CONF_FILE $INSTALL_DIR/
 
 RUN cd $INSTALL_DIR && \
-    yum -y install oracle-database-preinstall-18c file openssl $INSTALL_FILE_1 && \
+    yum -y install openssl oracle-database-preinstall-18c && \
+    yum -y install $INSTALL_FILE_1 && \
```
Djelibeybi: Why the split into two yum commands?

gvenzl: Because the post script of oracle-database-preinstall overwrites the memlock settings after the oracle-database rpm deactivates it and then `su` will no longer work and the install will fail as it can't switch to the `oracle` user. Sucks that post scripts aren't executed in order but just at the once all packages have been installed.

Issue #1231, opened 2019-04-15
typo in OracleFMWInfrastructure/dockerfiles/12.2.1.3/env.txt 


xiancao:

xiancao:
In OracleFMWInfrastructure/dockerfiles/12.2.1.3/env.txt file, it has:
ID=InfraDB
it should be:
DB_SID=InfraDB

otherwise the DB client can not connect to it using:
$ docker run -ti --network=InfraNET --rm store/oracle/database-instantclient:12.2.0.1 sqlplus sys/Oradoc_db1@InfraDB:1521/InfraDB.us.oracle.com AS SYSDBA

PR #1232, opened 2019-04-15
Clarify comment in run scripts, correct parameter name DB_SID, make h…

mriccell:
1- Correct parameter name DB_SID issue #1231 
2- Clarify comments about host directory in run scripts
3- Make host volume URL generic


OracleFMWInfrastructure/dockerfiles/12.2.1.3/run_admin_server.sh
```diff
@@ -18,7 +18,8 @@ set_context() {
 set_context
 . ${scriptDir}/container-scripts/setEnv.sh ${scriptDir}/properties/domain.properties
 
-hostvolume=/Users/mriccell/temp
+# HOST volume where the domain home will be persisted
+hostvolume=/Users/mydirectory/temp
```
Djelibeybi: I would suggest changing this to `/home/username/domain_home` as the current example is only used on macOS and I'd prefer to use a more Linux-specific example here. Also, the comment states "will be persisted" so the name "temp" seems incorrect to me.

OracleFMWInfrastructure/samples/12213-domain-in-volume/run_admin_server.sh
```diff
@@ -18,6 +18,7 @@ set_context() {
 set_context
 . ${scriptDir}/container-scripts/setEnv.sh ${scriptDir}/properties/domain.properties  ${scriptDir}/properties/rcu.properties
 
+# Host volume where the domain home will be persisted
 hostvolume=/Users/mydir/temp
```
Djelibeybi: I would suggest changing this to `/home/username/domain_home` as the current example is only used on macOS and I'd prefer to use a more Linux-specific example here. Also, the comment states "will be persisted" so the name "temp" seems incorrect to me.

mriccell: thanks for the suggestion, made changes

OracleFMWInfrastructure/dockerfiles/12.2.1.3/run_admin_server.sh
mriccell: thanks for the suggestion, made changes

mriccell:
The Domain_HOME will always be the DOMAIN_ROOT which is hard coded and appended by the DOMAIN_NAME.  The user passes the DOMAIN_NAME into the properties otherwise it uses the default value.

Djelibeybi:
I got myself confused again. I really shouldn't review stuff before 8am. :) I meant the `hostvolume` parameter (which contains the text `domain_home` which is where my confusion started). You're hard-coding it inside a script that will be overwritten by a `git pull` so I suggest moving it out into the environment variables file like the other parameters.

mriccell:
Oh ok got it!

mriccell:
Hi Avi do you mind reviewing the change?

OracleFMWInfrastructure/dockerfiles/12.2.1.3/run_admin_server.sh
```diff
@@ -19,19 +18,21 @@ set_context
 . ${scriptDir}/container-scripts/setEnv.sh ${scriptDir}/properties/domain.properties
 
 # HOST volume where the domain home will be persisted
-hostvolume=/home/username/domain_home
-echo "Host volume to write the domain is: $hostvolume"
+domain_host_volume() {
+
+   domainhostvol=${DOMAIN_HOST_VOLUME:-"/home/username/domain_home"}
```
Djelibeybi: There doesn't appear to be a test to see if  ${DOMAIN_HOST_VOLUME} exists and the default of `/home/username/domain_home` is almost certainly _not_ going to exist. This seems (to me) to be the kind of option that shouldn't have a default because we have no idea where/how the user is going to run the script and should rather test for the existence of ${DOMAIN_HOST_VOLUME} and error out if it doesn't exist instead of trying to create a Docker container with a bind mount to something broken.

mriccell: The DOMAIN_HOME_VOLUME gets created when the container gets run if it does not already exist.  I want to reuse it if it was created in a previous run, that is if the admin server container comes down and I re-start I do not want the DOMAIN_HOME_VOLUME created again.  If DOMAIN_HOME_VOLUME is an invalid directory the container will fail to run, WebLogic will not configure the domain and it will throw an error. 
I will make changes to not take a default and make sure that the DOMAIN_HOME_VOLUME is set in the domain.properties file.

Djelibeybi: Sounds good to me. :)

PR #1241, opened 2019-04-19
Coherence 12.2.1.3.2 OPatch example

fryp:
Add instructions for patching Coherence 12.2.1.3 image to Coherence 12.2.1.3.2 bundle patch


edburns:
Patrick can you please state that the location of `opatch` is `/u01/oracle/oracle_home/OPatch`?

Djelibeybi:
This sample is specific to a particular patch and doesn't provide any instructions on how/why/where to change which bits for a different patch. Might want to either make the sample generic i.e. it takes a patch number as a parameter to `buildImage.sh` or you'll want more documentation around how to update the `Dockerfile`, etc.

fryp:
Will be switching to `dockerfiles/12.2.1.3.2`.  Closing this PR.

PR #1243, opened 2019-04-19
Improvments to the 12213-domain-in-volume scripts to check for DOMAIN…

mriccell:
Add check that the DOMAIN_HOST_VOLUME is a valid directory and that it is set
Add check in run_managed_server that the server name and mapped port are passed into the script

### DIFF
--- a/OracleFMWInfrastructure/samples/12213-domain-in-volume/container-scripts/createFMWDomain.py
+++ b/OracleFMWInfrastructure/samples/12213-domain-in-volume/container-scripts/createFMWDomain.py
@@ -212,9 +212,9 @@ def usage():
     sys.exit(0)
 
 # Uncomment for Debug only
-print str(sys.argv[0]) + " called with the following sys.argv array:"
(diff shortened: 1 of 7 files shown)


OracleFMWInfrastructure/samples/12213-domain-in-volume/container-scripts/createFMWDomain.py
```diff
@@ -212,9 +212,9 @@ def usage():
     sys.exit(0)
 
 # Uncomment for Debug only
-print str(sys.argv[0]) + " called with the following sys.argv array:"
-for index, arg in enumerate(sys.argv):
-    print "sys.argv[" + str(index) + "] = " + str(sys.argv[index])
+#print str(sys.argv[0]) + " called with the following sys.argv array:"
```
Djelibeybi: Don't commit commented out code: remove it. :) It'll always be in your git history if you need to refer to it later.

Issue #1292, opened 2019-05-21
libaio package missing with latest 12.2.1.3 OracleFMWInfrastructure dockerfiles


ashageetha:
**Issue**:
When we built OracleBI 12.2.1.3 image with latest 12.2.1.3 OracleFMWInfrastructure dockerfiles, 
 system components did not start as libnqcsfwrapper.so could not load. This was because of dependent libaio.so.1 was not present.

**Analysis**:
After debug we found that initially libaio which was being shipped with OracleFMWInfrastructure  12.2.1.3 image is missing with latest dockerfiles. 

**Root Cause**:
When multistage build was implemented to  Shrink the image size, install of libaio was done in intermediate first build and was not installed with final build. Hence libaio is missing in final 12.2.1.3 OracleFMWInfrastructure image.

**Dockerfile details**:
libaio installed in intermediate step of build:  https://github.com/oracle/docker-images/blob/1714789336409f66b4027ee1c11a860ad13d196b/OracleFMWInfrastructure/dockerfiles/12.2.1.3/Dockerfile#L59
But not present with final build: https://github.com/oracle/docker-images/blob/1714789336409f66b4027ee1c11a860ad13d196b/OracleFMWInfrastructure/dockerfiles/12.2.1.3/Dockerfile#L88

Djelibeybi:
I believe @mriccell is already aware of this.

Djelibeybi:
Fixed by #1293.

PR #1293, opened 2019-05-21
Fix OracleFMWInfrastructure image

mriccell:
1) Improve Dockerfile to make image smaller
2) Fix install of libaio library
3) Remove VOLUME directive
4) Correct README to indicate to use -v to map domain home to host volume


mriccell:
Fixed #1292

Djelibeybi:
Approved, but I suggest updating line 100 of the README to include an actual example usage of the `-v` option (for people who don't use our scripts or use an automation/scheduling engine).

PR #1297, opened 2019-05-23
Separate check of existence and checksum check for packages for OracleDatabase

McNetic:
This way, the existence of the required files will always be checked, even if checksum check is skipped. This should, of course, be applied to all other build scripts as well..

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
@@ -35,12 +35,22 @@ EOF
 }
 
 # Validate packages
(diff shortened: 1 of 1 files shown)


Djelibeybi:
Thanks for your submission. Could you please confirm whether you've signed the OCA by signing off your commit with the email address used to submit your OCA at https://www.oracle.com/technetwork/community/oca-486395.html.

If not, please follow the process on that page to submit your OCA. Once that's done, we can merge your code.

Djelibeybi:
Tagging this with more-info-required due to lack of response.

Djelibeybi:
Closed due to lack of response.

PR #1344, opened 2019-08-04
Makes README gender neutral with removal of 'he' and 'his'

cdavid15:
Current README implies male users with use of 'he' and 'his' which is completely unnecessary.

Simple change of wording to make it gender neutral.


Djelibeybi:
What a great change. Thanks for this.

cdavid15:
Decided to check the rest of the repository and found a few more references which I have fixed in the commit I just pushed.

OracleDatabase/RAC/OracleRealApplicationClusters/dockerfiles/12.2.0.1/sshUserSetup.sh
```diff
@@ -16,7 +16,7 @@
 # confirmations.
 #Please note that the script would remove write permissions on the remote hosts
 #for the user home directory and ~/.ssh directory for "group" and "others". This
-# is an SSH requirement. The user would be explicitly informed about this by teh script and prompted to continue. In case the user presses no, the script would exit. In case the user does not want to be prompted, he can use -confirm option.
+# is an SSH requirement. The user would be explicitly informed about this by teh script and prompted to continue. In case the user presses no, the script would exit. In case the user does not want to be prompted, they can use the -confirm option.
```
rjeberhard: teh -> the

rjeberhard: We should probably run the whole file past one of our doc. writers.  "In case the user does not want to be prompted, they can use the -confirm option." might be better worded as "Use the -confirm option to skip confirmation prompts."  Then we also don't have the singular/plural confusion to work around being sensitive to gender.

Djelibeybi: And the `-confirm` option should be formatted as monotype using back ticks.

PR #1358, opened 2019-08-22
Resolve issues with the WebLogic domain-in-image samples when patches are applied on top of the domain image

TheFrogPad:
WebLogic Operator team has gotten reports from other Oracle teams that the samples used to create a domain home in a docker image did not function when an OPatch was applied to on the docker image created with the domain home.

Updates to the WebLogic samples are required such that the images run as `oracle` user which then enables an OPatch to be applied on top of the domain image. File permission errors resulted without these changes to how the Dockerfiles setup directories an apply the OPatch. An additional update to the docker build flags was made so that the docker cache is not used.

### DIFF
--- a/OracleFMWInfrastructure/samples/12213-domain-home-in-image/Dockerfile
+++ b/OracleFMWInfrastructure/samples/12213-domain-home-in-image/Dockerfile
@@ -67,11 +67,11 @@ COPY container-scripts/* /u01/oracle/container-scripts/
 #Create directory where domain will be written to
 USER root
 RUN chmod +x /u01/oracle/container-scripts/* && \
-    mkdir -p ${PROPERTIES_FILE_DIR} && \
+    mkdir -p +xwr ${PROPERTIES_FILE_DIR} && \
     chown -R oracle:oracle ${PROPERTIES_FILE_DIR} && \
     mkdir -p $DOMAIN_HOME && \
-    chown -R oracle:oracle $DOMAIN_HOME && \
-    chmod -R a+xwr $DOMAIN_HOME
+    chown -R oracle:oracle $DOMAIN_HOME/.. && \
+    chmod -R a+xwr $DOMAIN_HOME/..
 
 USER oracle
 COPY properties/*.properties ${PROPERTIES_FILE_DIR}/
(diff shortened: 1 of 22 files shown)


TheFrogPad:
@ddsharpe @markxnelson Can you please review?

OracleFMWInfrastructure/samples/12213-domain-home-in-image/Dockerfile
```diff
@@ -67,12 +67,13 @@ COPY container-scripts/* /u01/oracle/container-scripts/
...
 USER root
 RUN chmod +x /u01/oracle/container-scripts/* && \
-    mkdir -p ${PROPERTIES_FILE_DIR} && \
+    mkdir -p +xwr ${PROPERTIES_FILE_DIR} && \
     chown -R oracle:oracle ${PROPERTIES_FILE_DIR} && \
     mkdir -p $DOMAIN_HOME && \
-    chown -R oracle:oracle $DOMAIN_HOME && \
-    chmod -R a+xwr $DOMAIN_HOME
+    chown -R oracle:oracle $DOMAIN_HOME/.. && \
+    chmod -R a+xwr $DOMAIN_HOME/..
```
ddsharpe: I don't think this was the intention of the writer (of the sample).  I think the intention was to change the permissions of the domain home directory.  Why is it necessary to change the ownership of its parent directory?

TheFrogPad: The sample will not work without changing the parent folder of the domain, the create domain handling in WLST fails with file permission errors.

Also, the `domain-home-in-image-wdt` sample which is switching back to `oracle` user from `root addresses the file permissions from the domain parent folder level. These changes are as such consistent between these domain in image samples.

One additional comment from me is that since `oracle` is the user that will run the image, having the parent folder ownership is quite valid IMHO since the `oracle` user can create/use additional folders for `user_projects` like logs or even additional domain folders can be created.

TheFrogPad:
Given some feedback it would appear that the `COPY` command does not run as the `USER` so I have made change to ensure all the files being copied are owned by `oracle` user.

I've also addressed @markxnelson comment to note the use of the `--no-cache` options.

TheFrogPad:
@mriccell this change is ready pending your review and approval... Thanks!

PR #1364, opened 2019-08-27
Update createWLSDomain.sh

tabeth:
Reference Issue: #1340 

Prevent the existence of the file from resulting in an exit. The implication is that with the previous line, restarting the Docker container would immediately exit as the first run of this would create the log file this is checking the existence for.

### DIFF
--- a/OracleWebLogic/samples/12213-domain/container-scripts/createWLSDomain.sh
+++ b/OracleWebLogic/samples/12213-domain/container-scripts/createWLSDomain.sh
@@ -30,7 +30,8 @@ trap _kill SIGKILL
 export DOMAIN_HOME=$CUSTOM_DOMAIN_ROOT/$CUSTOM_DOMAIN_NAME
 echo "Domain Home is:  $DOMAIN_HOME"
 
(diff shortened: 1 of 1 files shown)


tabeth:
@mriccell 
> This is not the right fix because if you exit if there is no log file then the very first time that the domain needs to get created it will not. The fix should be if there already is a log file do not create the domain.

You're right. The fix I suggested only happened to work since I had built the image already once. Per your suggestion we'd want something like this:

```bash
if [ ! -f ${DOMAIN_HOME}/servers/${CUSTOM_ADMIN_NAME}/logs/${CUSTOM_ADMIN_NAME}.log ]; then
    echo "Log file cannot be found. Creating..."
    wlst.sh -skipWLSModuleScanning -loadProperties ${DOMAIN_PROPERTIES_FILE} -loadProperties ${SEC_PROPERTIES_FILE}  /u01/oracle/container-scripts/create-wls-domain.py
retval=$?
fi

-- (existing script)...

if [ ! -f ${DOMAIN_HOME}/servers/${CUSTOM_ADMIN_NAME}/logs/${CUSTOM_ADMIN_NAME}.log ]; then
     touch ${DOMAIN_HOME}/servers/${CUSTOM_ADMIN_NAME}/logs/${CUSTOM_ADMIN_NAME}.log
fi
```

(this could probably also be done in the same condition for brevity).

PR #1368, opened 2019-09-02
Added v19 of ORDS and listening to any interfaces

malagoli:
Hi,
added the checksum for latest version of ords and modify the configuration file to listen on any address, this allow not to specify the ip address in a k8s environment


Djelibeybi:
You seem to have combined two changes into a single pull request. I'm seeing changes to both the RAC Dockerfiles/scripts as well as the ORDS ones. Also, you have switched one of the source images to the container-registry, which we don't do here on GitHub.

I'm going to close this PR as it's a little too messy to merge. I would suggest creating a separate branch from master for each changeset so that each PR can be reviewed just by a single codeowner.

You can learn more about the standard GitHub Flow here: https://guides.github.com/introduction/flow/

PR #1384, opened 2019-09-27
Oracle GoldenGate on Docker

sbalousek:
- Change conditional logic from '! -z' to clearer '-n'
 - Do not apply the default tag if the user specified a tag on the command line
 - Tested with OGG 19.1.0.0.2


OracleGoldenGate/dockerBuild.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2017 Oracle and/or its affiliates.  All rights reserved.
+# Copyright (c) 2017, 2019 Oracle and/or its affiliates.  All rights reserved.
```
Djelibeybi: This should just be 2019.

PR #1390, opened 2019-09-30
1373 fixes

rishabh20oracle:
The following changes have been done in this pull request which corresponds to the changes proposed in Issue # 1373:
1. Created a folder generic_build_scripts under dockerfiles folder using the files present under dockerfiles/19.3.0 folder. This can be used to create docker images starting from 20c.
2. Modified files dbca.rsp.tmpl and db_inst.rsp under generic_build_scripts folder to remove hardcoded major version (18, 19, ...) and replaced it with a placeholder {MAJOR_VERSION}.
3. Modified Dockerfile inside generic_build_scripts folder to accept 2 arguments MAJOR_VERSION and GOLDIMAGE_NAME so as to make docker build generic across versions starting from 20c.
4. Modified buildDockerImage.sh to change directory to generic_build_scripts when the passed version is 20 or higher.
5. Removed Checksum.ee and Checksum.se2 from generic_build_scripts and added a new file Checksum which will host the checksums of all the new goldimages (starting from 20c). This file will be updated as and when a new goldimage is uploaded to OTN.


latakannan:
_
Just to clarify on this.. What Rishabh meant below is the on prem release and not the cloud release. So the effort here is to unify the scripts and see if one could avoid creating new directories for every on prem release that is made for the database.  
> Generic building of DB Single Instance docker images starting from 20c instead of adding a new directory every time a new cloud release happens. 12 hours ago_

Djelibeybi:
This pull request has been abandoned for over a year. Closing.

PR #1403, opened 2019-10-14
Enable SSL support for 12.2.1.3 domain home in image sample

qvdang:
Hi Monica,
Please review this change that supports enabling SSL in the WLS servers for the domain home in image sample.  I have verified the change by running the WLS instances in the individual docker containers and also as outlined in the WLS K8S Operator QuickStart Guide simple sample(https://oracle.github.io/weblogic-kubernetes-operator/samples/simple/domains/domain-home-in-image/).  
@TheFrogPad also please take a look.
Thanks.


OracleWebLogic/samples/12213-domain-home-in-image/container-scripts/startManagedServer.sh
```diff
@@ -39,11 +39,14 @@ if [ -z "${PASS}" ]; then
 fi
 
 #Set Java Options
-JAVA_OPTIONS=`awk '{print $1}' ${SEC_PROPERTIES_FILE} | grep ^JAVA_OPTIONS= | cut -d "=" -f2`
-if [ -z "${JAVA_OPTIONS}" ]; then 
-   JAVA_OPTIONS="-Dweblogic.StdoutDebugEnabled=false"
+# Use the Env JAVA_OPTIONS if it's been set
+if [ -z "$JAVA_OPTIONS" ]; then
```
TheFrogPad: IMHO it would be a better choice to have this logic augment the command line `JAVA_OPTIONS` value with the options specified in the properties file rather than simply have the command line override the properties setting?

I'm thinking of the case where the user does place some expected `JAVA_OPTIONS` in the properties file but then wants to supplement those with temporary debug flags for WebLogic or even JDK network SSL debug.

In the `README.md` the updated text does say override but we should be clear that it's a replacement or that the value set on the command line will augment as I'm suggesting that we do here...

qvdang:
Hi @mriccell , would you please view this PR?  The change is needed to support SSL enabling for WLS.  The initial SSL config uses the WLS Demo identity & trust.  After that I'll expand it to provide an example on how to use custom identity & trust possibly with config overrides.
Thanks!

OracleWebLogic/samples/12213-domain-home-in-image/README.md
```diff
@@ -73,17 +76,19 @@ To start the containerized Administration Server, run:
           -v <HOST DIRECTORY TO PROPERTIES FILE>/properties/docker-run:/u01/oracle/properties \
           12213-domain-home-in-image
 
+If both the JAVA_OPTIONS environment variable and the JAVA_OPTIONS in the docker-run/security.properties file are configured, the latter gets appended to the former. If SSL is enabled, pass "-e JAVA_OPTIONS=-Dweblogic.security.SSL.ignoreHostnameVerification=true" to the "docker run ..." command that starts a managed server because the Demo identity certificate got generated with a host name different from the admin host above when the domain was created.
```
mriccell: Instead of "command that starts a managed server  because the Demo identity certificate got generated with a host name different from the admin host above when the domain was created." it reads better "command that starts a managed server.  This is necessary because the Demo identity certificate gets generated with a host name different from the admin server host when the domain is created."

mriccell: These changes should also be made to the sample domain-home-in-image-wdt

qvdang: @mriccell , I've updated the README.md as suggested.  I also made the same changes to enable SSL in the WDT sample.  Please take a look. Thanks!

PR #1406, opened 2019-10-16
Read JAVA_OPTIONS fix

dandresm:
If you add JVM arguments of type -Dkey1=value1 -D key2=value2... in JAVA_OPTIONS property of file security.properties, only the "-Dkey1" part gets captured and added as JVM arguments when starting the server. Adding the "-" to the "-f2" argument fix the problem because you get the rest of the value. This could also be applyed to password value because equals sign could be part of a password.


rjeberhard:
@mriccell, you need to please look at this too.

PR #1408, opened 2019-10-17
Updated the docker file to match what we do with the JDK 11 docker file

aureliogrb:
This is now closer to what is done on OpenJDK makefiles


aureliogrb:
jshell was introduced with JDK 9. It is not available with JDK/ServerJRE 8

aureliogrb:
> 
> 
> Please add `CMD["jshell"]` to match the OpenJDK image.
jshell was introduced with JDK 9. It is not available with JDK/ServerJRE 8

PR #1418, opened 2019-10-28
Correct permissions in 12.2.1.4 images

mriccell:
Correct permissions to /u01 in Dockerfiles
Add statement that clarifies that we support multi-server as well as single-server domain.

### DIFF
--- a/OracleWebLogic/dockerfiles/12.2.1.4/Dockerfile.generic
+++ b/OracleWebLogic/dockerfiles/12.2.1.4/Dockerfile.generic
@@ -52,7 +52,7 @@ ENV FMW_PKG=fmw_12.2.1.4.0_wls_Disk1_1of1.zip \
 
 # Copy packages
 # -------------
-COPY $FMW_PKG install.file oraInst.loc rda_lin64 rda.cfg /u01/
+COPY $FMW_PKG install.file oraInst.loc /u01/
 RUN chown oracle:oracle -R /u01
 
 # Install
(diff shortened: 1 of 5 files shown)


mriccell:
The correction to the permissions to /u01 fixes problem seen with WebLogic Kubernetes Operator "DOMAIN INTROSPECTOR LEAKS INFORMATION FROM OTHER DOMAINS"

OracleWebLogic/dockerfiles/12.2.1.4/Dockerfile.generic
```diff
@@ -94,6 +94,7 @@ COPY --from=builder --chown=oracle:oracle /u01 /u01
 COPY container-scripts/createAndStartEmptyDomain.sh container-scripts/create-wls-domain.py /u01/oracle/
 
 RUN chmod +xr $SCRIPT_FILE && \
+    chown oracle:oracle -R /u01 && \
     chown oracle:oracle -R /u01/oracle/createAndStartEmptyDomain.sh && \
```
rjeberhard: `chown -R` means apply recursively, so if you run at /u01, then you no longer need the following lines specifically for the scripts.  Could we do this at /u01/oracle?

mriccell: the problem is that the Oracle Home is at /u01 and the problem with the introspection job is due to permissions to the /u01 directory.  Do you mean replace the two scripts chown statements with one "chown oracle:oracle -R /u01/oracle" or to remove all together the chown statement for the scripts?

rjeberhard: I was asking two different things: a) If `chown -R /u01/oracle` was sufficient?, and b) either way, once we have `chown -R /u01` or `chown -R /u01/oracle` then the other lines in the same statement are unnecessary since those paths are under /u01/oracle.

sbattagi:
Just observed a typo at line 27.
# From the OIracle Docker GitHub Project

mriccell:
> Just observed a typo at line 27.
> # From the OIracle Docker GitHub Project

Which file?

sbattagi:
> 
> 
> > Just observed a typo at line 27.
> > # From the OIracle Docker GitHub Project
> 
> Which file?

Sorry, I thought I was commenting on the file. 
This is the file:  https://github.com/oracle/docker-images/blob/master/OracleWebLogic/dockerfiles/12.2.1.4/Dockerfile.generic

PR #1434, opened 2019-11-22
Fix typo in the RAC storage Server runOracle.sh files

fojogit:
Change 'Assining' to 'Assigning'


Djelibeybi:
Thank you for this, well-spotted.

Issue #1447, opened 2019-12-03
Multiple hosts and clustered weblogic servers


acarlson1994:
The 1221-multihost samples folder has been removed from the weblogic docker-images folder. See this issue:
https://github.com/oracle/docker-images/issues/52#issue-123165182
And related commit:
https://github.com/oracle/docker-images/commit/6184711c110a02468346a611c55db5e41145595d

Have these files been removed because multihost is not supported? As it is stated in the documentation:
https://docs.oracle.com/middleware/1221/wls/DOCKR/configuration.htm#DOCKR148

Both the issue and commit referenced above do not explain enough IMO.

Is there any plan for supporting multihost?
Or does multihost work relatively fine, but just isnt officially supported?

mriccell:
Multi host is supported! We have evolved WebLogic support to run Docker containers in Kubernetes please see https://github.com/oracle/weblogic-kubernetes-operator.  This is an old sample and WebLogic 12.2.1 is no longer supported, I will cleanup all references to this sample.

mriccell:
@acarlson1994 any other questions?

acarlson1994:
@mriccell no questions. Thank you for the quick response / action!

PR #1448, opened 2019-12-04
removing outdated workshops

mriccell:
removing outdated workshops


mriccell:
Removing workshops for multi host which is outdated #1447

Issue #1491, opened 2020-01-14
Reduce the size of Oracle Coherence 12.2.1.4.0 docker image


dhirupandey:
Changes to reduce the size of Oracle Coherence 12.2.1.4.0 docker image caused by the following in the DockerFile

```
RUN chgrp -R 0 /u01 && \
    chmod -R g=u /u01 && \
    chmod 755 /start.sh && \
    chgrp -R 0 /start.sh && \
    chmod -R g=u /start.sh && \
    chgrp -R 0 /logs && \
    chmod -R g=u /logs
```
Moved some of this to the builder stage in DockerFile to reduce the size. See the resulting size from the released version on OCR, to the current version and then with the changes:

```
12.2.1.4.0>docker images | grep 12.2.1.4
oracle/coherence                          12.2.1.4.0              111ad2e768fa        10 minutes ago      385MB
oracle/coherence                          12.2.1.4.0-current      89aefbe2bed2        2 days ago          491MB
container-registry.oracle.com/middleware/coherence   12.2.1.4.0              70a26e1b991c        3 months ago        583MB
12.2.1.4.0>
```

dhirupandey:
Fixed by PR https://github.com/oracle/docker-images/pull/1492

PR #1492, opened 2020-01-14
Reduce the size of Oracle Coherence 12.2.1.4.0 docker image

dhirupandey:
https://github.com/oracle/docker-images/issues/1491


dhirupandey:
> Change lines 95-99 with the following:
> 
> ```
> mkdir -p /u01/oracle && \
> chgrp -R root /u01 && \
> chmod -R g=u /u01 && \
> ```
> 
> You don't need to repeat it over and over. Just create the entire directory path and then set the ownership and permissions recursively from `/u01`. I'd also consider moving `/logs` to `/u01/logs` so that it's also covered by the recursion but also to move it out of /.

Good point ... let me make that change

dhirupandey:
Not sure about the /logs directory ... maybe that is required for running on openshift, so I won't change that. Perhaps this is something that @thegridman can decide

Djelibeybi:
> Not sure about the /logs directory ... maybe that is required for running on openshift, so I won't change that. Perhaps this is something that @thegridman can decide

Good point. I'm not that familiar with OpenShift's requirement here either.

PR #1499, opened 2020-01-20
Changes for spike troubleshooting

davidkelliott:
We had to remove the OSB component otherwise the image size exceeded 10GB and wouldn't upload to ECR.

To build the image you need to log into and accept the terms for the oracle images that this image is based off.  See the original read me, but here are some extra steps that are not covered in it:
https://container-registry.oracle.com/
Then log into their repository with docker - https://docs.oracle.com/cd/E37670_01/E75728/html/oracle-registry-server.html
And download the required images:
oracle/fmw-infrastructure:12.2.1.3
The images download with different names, tag them with the name above.
You also need to have the SOA binaries fmw_12.2.1.3.0_soa.jar present in the 12.2.1.3 folder.
Build the image with `docker-images/OracleSOASuite/dockerfiles/buildDockerImage.sh -v 12.2.1.3`
Then tag it to match the name of the SOA image in the ECR repository


davidkelliott:
Sorry opened in error

Issue #1443, opened 2019-11-29
unable to use wlst.sh connect() method on soa admin server


ddevadat:
i have started soa admin server using the below command

docker run -i -t  --name soaas --network=SOANet -p 7001:7001  -v /scratch/DockerVolume/SOAVolume/SOA:/u01/oracle/user_projects   --env-file ./adminserver.env.list oracle/soa:12.2.1.3

Admin server is up and am able to access em console
But i am not able to connect via wlst.sh connect method.

I tried all the below method, none of them working
  connect('weblogic','Welcome1','localhost:7001')
  connect('weblogic','Welcome1','ad15c2a154c2:7001')
  connect('weblogic','Welcome1','172.18.0.3:7001')
  connect('weblogic','Welcome1','soaas:7001')
  connect('weblogic','Welcome1','0.0.0.0:7001')

Getting 
javax.naming.CommunicationException: Failed to initialize JNDI context, tried 2 time or times totally, the interval of each time is 0ms.
Timed out while attempting to establish connection to :t3://ad15c2a154c2:7001 [Root exception is java.io.IOException: Timed out while attempting to establish connection to :t3://ad15c2a154c2:7001]

ddevadat:
solved this by updating external dns address for admin server to 172.18.0.3 from the default InfraAdminContainer

PR #1501, opened 2020-01-22
SOA Readme update to add ADMIN_HOST in environment file adminserver.env.list

ashageetha:
README.md does not gives details on adding **ADMIN_HOST** in environment file "adminserver.env.list ".

If ADMIN_HOST value is not passed during starting Admin Server container,  it will wrongly pick  ADMIN_HOST value as **InfraAdminContainer**, default value of base image (oracle/fmw-infrastructure:12.2.1.3) i.e., from https://github.com/oracle/docker-images/blob/c5b1cbb04af0323ef8495becb6ad5b5481311f1e/OracleFMWInfrastructure/dockerfiles/12.2.1.3/Dockerfile#L66

If ADMIN_HOST not set to correct value, WebLogic Console will not be accessible as this value will be updated external dns address ( as reported in #1443 )

### DIFF
--- a/OracleSOASuite/dockerfiles/README.md
+++ b/OracleSOASuite/dockerfiles/README.md
@@ -125,6 +125,7 @@ Create an environment file adminserver.env.list
     ADMIN_PASSWORD=<admin_password>
     MANAGED_SERVER=<managed servername>
     DOMAIN_TYPE=<soa/osb/bpm>
+    ADMIN_HOST=<Admin host name>
     
 Sample Data will look like this:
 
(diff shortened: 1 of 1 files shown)


OracleSOASuite/dockerfiles/README.md
```diff
@@ -135,6 +136,7 @@ Sample Data will look like this:
     ADMIN_PASSWORD=Welcome1
     MANAGED_SERVER=soa_server1
     DOMAIN_TYPE=soa
+    ADMIN_HOST=slc09cwi.us.oracle.com
```
mriccell: You are adding your host address please replace for a dummy ADMIN_HOST

ashageetha: Sure. I was reusing the same code which was already available for soaserver.env.list for Managed Server. I'll update at both places and push the changes.

Djelibeybi:
This pull request has merge conflicts and needs to be rebased against master.

Djelibeybi:
Fixed via #1512

PR #1512, opened 2020-01-28
SOA 12.2.1.3 Dockerfile updated for multipart build

sbattagi:


OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
+SOA on Docker
+=============
+
+Sample Docker configurations to facilitate installation, configuration, and environment setup for Docker users. This project includes quick start dockerfiles for Oracle SOA 12.2.1.3.0 based on Oracle Linux 7, Oracle JRE 8 (Server) and Oracle Fusion Middleware Infrastructure 12.2.1.3.0.
+
+At the end of this configuration there may be 3 containers running: 
+1. Oracle Database Container (Only when container based Database is used)
```
mriccell: (Optional: only when RCU schema is created in a database running in a container)

I would add optional just to make it clear

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+    DB_SCHEMA_PASSWORD=Welcome1
+    ADMIN_PASSWORD=Welcome1
+    MANAGED_SERVER=soa_server1
+    DOMAIN_TYPE=soa
+    
+To start a docker container with a SOA domain and the WebLogic Administration Server call `docker run` command and pass the above `adminserver.env.list` file.
+
+A sample docker run command is given below:
+
+     $docker run -i -t  --name soaas --network=SOANet -p 7001:7001  -v /scratch/DockerVolume/SOAVolume/SOA:/u01/oracle/user_projects   --env-file ./adminserver.env.list oracle/soa:12.2.1.3.0
```
mriccell: $ docker
provide a blank space after $ sign

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+    
+To start a docker container for SOA server you can simply call `docker run` command and passing `soaserver.env.list`. 
+
+A sample docker run command is given below:
+
+    $ docker run -i -t  --name soams --network=SOANet -p 8001:8001   --volumes-from soaas   --env-file ./soaserver.env.list oracle/soa:12.2.1.3.0 "/u01/oracle/dockertools/startMS.sh"
+    
+Using `--volumes-from` reuses the volume created by the Administration Server container. In the above `docker run` command, `soaas` is the name of the Administration Server container started in the previous step and we must use the same name used for the Administration Server to start the SOA Managed Server. 
+
+This includes the command to tail logs to keep the container up and running.
```
mriccell: I think this sentence is not necessary.  You could instead provide the command so that a user can look at the logs of the Admin Server.

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+>IMPORTANT: You need to wait till all the above commands are run before you can start the SOA Server.
+
+The following lines highlight when the SOA Managed Server is ready to be used:    
+`INFO: Managed Server is running`
+
+`INFO: Managed server has been started`
+
+Once the SOA container is created logs will be tailed and displayed to keep the container running.
```
mriccell: Provide the command so that a user can tail the logs.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch
```diff
@@ -0,0 +1,81 @@
+#
```
mriccell: In our projects the patch is a separate sample with its own README for example https://github.com/oracle/docker-images/tree/master/OracleFMWInfrastructure/samples/12213-patch

OracleSOASuite/dockerfiles/README.md
```diff
@@ -1,102 +1,37 @@
 SOA on Docker
```
mriccell: This README should be very generic and not go into details of how to run the Admin Server and managed servers.  Each version of SOA 12.2.1.3 and 12.2.1.4 will have its own README. Look for example https://github.com/oracle/docker-images/tree/master/OracleFMWInfrastructure/README.md

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile
```diff
@@ -14,11 +14,12 @@
...
 # Pull base image
 # ---------------
-FROM oracle/fmw-infrastructure:12.2.1.3
+FROM oracle/fmw-infrastructure:12.2.1.3 as builder
+
 #
 # Maintainer
 # ----------
-MAINTAINER Prashanth Nagaraja <prashanth.nagaraja@oracle.com>
+LABEL maintainer="Prashanth Nagaraja <prashanth.nagaraja@oracle.com>"
```
Djelibeybi: Is Prashanth still the maintainer or are you? Either way, the `builder` doesn't need a maintainer label.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile
```diff
@@ -31,20 +32,36 @@ ENV FMW_JAR1=fmw_12.2.1.3.0_soa.jar \
...
   $JAVA_HOME/bin/java -jar $FMW_JAR1 -silent -responseFile /u01/soasuite.response -invPtrLoc /u01/oraInst.loc -jreLoc $JAVA_HOME -ignoreSysPrereqs -force -novalidation ORACLE_HOME=$ORACLE_HOME && \
   $JAVA_HOME/bin/java -jar $FMW_JAR2 -silent -responseFile /u01/soasuite.response -invPtrLoc /u01/oraInst.loc -jreLoc $JAVA_HOME -ignoreSysPrereqs -force -novalidation ORACLE_HOME=$ORACLE_HOME INSTALL_TYPE="Service Bus" && \
-  rm -fr /u01/*.jar /u01/*.response /u01/*.loc
+  rm -fr /u01/*.jar /u01/*.response
+
+#
+# Rebuild from base image
+# -----------------------
+FROM oracle/fmw-infrastructure:12.2.1.3
+
```
Djelibeybi: Move the maintainer label here.

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Mount a host directory as a data volume
+
+Data volumes are designed to persist data, independent of the container’s lifecycle. The default location of the volume in container is under `/var/lib/docker/volumes`. There is an option to mount a directory from the host into a container as volume. In this project we will use that option for the data volume. The volume will be used to store Database datafiles and WebLogic Server domain files. This volume will be created on the host at `/scratch/DockerVolume/SOAVolume/`. Since the volume is created as "root" user, provide read/write/execute permissions to "oracle" user (by providing permissions to "others"), as all operations inside the container happens with "oracle" user login.
+
+To determine if a user already exists on your host system with uid:gid of 1000, run:
+
+    # getent passwd 1000
+
+If that returns a username (which is the first field), use that user for the `useradd` command below. If not, create the `oracle` user manually.
```
Djelibeybi: If `getent` returns a username, then a user with a `uid` of `1000` already exists and any attempt to run `useradd` with the same name should fail or cause inconsistent results (depending on the host OS).

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+    # getent passwd 1000
+
+If that returns a username (which is the first field), use that user for the `useradd` command below. If not, create the `oracle` user manually. 
+
+    # useradd -u 1000 -g 1000 oracle 
+
+Run the following commands as root:
+
+    # mkdir -p /scratch/DockerVolume/SOAVolume/
```
Djelibeybi: Please don't promote the use of `/scratch` as a valid directory in which to store data. It's not part of the [filesystem hierarchy standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). In this instance, as you are installing to `/u01` inside the container, I would suggest using it as the example outside the container as well.

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+    MANAGED_SERVER=soa_server1
+    DOMAIN_TYPE=soa
+    
+To start a docker container with a SOA domain and the WebLogic Administration Server call `docker run` command and pass the above `adminserver.env.list` file.
+
+A sample docker run command is given below:
+
+     $docker run -i -t  --name soaas --network=SOANet -p 7001:7001  -v /scratch/DockerVolume/SOAVolume/SOA:/u01/oracle/user_projects   --env-file ./adminserver.env.list oracle/soa:12.2.1.3.0
+     
+>IMPORTANT: the resulting images will NOT have a domain pre-configured. But, it has the scripts to create and configure a soa domain >while creating a container out of the image.
```
Djelibeybi: The formatting on this line is still broken. You need a LF before the second `>`

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+
+## Creating a container for SOA Server
+
+Start a container to launch the Managed Server from the image created. The environment variables used to run the Managed Server image are defined in the file `soaserver.env.list`. 
+
+Create an environment file `soaserver.env.list` with the below details:
+
+    MANAGED_SERVER=<Managed Server name, For Exp:- soa_server1, osb_server1, bpm_server1>
+    DOMAIN_TYPE=<soa/osb/bpm>
```
Djelibeybi: What's the difference between the three? Which one should I pick? What do they mean?

OracleSOASuite/dockerfiles/12.2.1.3/README.md
```diff
@@ -0,0 +1,170 @@
...
+    DOMAIN_TYPE=soa
+    ADMIN_HOST=host.domain.com
+    ADMIN_PORT=7001
+    ADMIN_PASSWORD=Welcome1
+    
+To start a docker container for SOA server you can simply call `docker run` command and passing `soaserver.env.list`. 
+
+A sample docker run command is given below:
+
+    $ docker run -i -t  --name soams --network=SOANet -p 8001:8001   --volumes-from soaas   --env-file ./soaserver.env.list oracle/soa:12.2.1.3.0 "/u01/oracle/dockertools/startMS.sh"
```
Djelibeybi: Why does the user need to provide a command here? And why is it wrapped in quotes? The container should just know what to run, you shouldn't have to rely on a customer getting this right. Especially when it may be driven by automation or scaling.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch
rdas76: My understanding is if we use the Docker files as in the patch sample it will create additional layers and will cause large image size. This is similar to the update option in the image tool. 

We agreed to create new images with original shiphomes and patches combined to optimize the size. The sample Docker file that we have created Dockerfile.patch achieves that purpose. 

Also keeping in a separate folder will be maintenance headache as we would like to ensure the Dockerfile and Dockerfile.patch should be exactly same except additional step fo applying the patch.

Hence, we decided to keep in the same folder. 

Let me know if you have any disagreement with the above plan.

OracleSOASuite/dockerfiles/README.md
rdas76: This README was created when we released 12.2.1.3 Docker for Dev and Test purpose. This  was written keeping in mind for standalone Docker deployment not with Kubernetes. Since, standalone Docker deployment still being supported, starting Admin and Managed servers instructions are still relevant and in scope. Let me if this is not correct.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile
rdas76: Samba, add your name as maintainer not Nagaraj as he is not involved in this project.

OracleSOASuite/dockerfiles/12.2.1.3/README.md
rdas76: These are 3 different applications running as containers from the same image. This is runtime setting not a build time setting. You just create one single SOA image and users can start different containers based on this environment settings. Hope this clarifies.

OracleSOASuite/dockerfiles/12.2.1.3/README.md
rdas76: This is very old README which was reviewed and approved already. This hasn't been changed since then. But anyways to clarify your question, by default container starts the Admin server and then we ask user to manually start the required Managed servers. This method of startup already reviewed and approved when we released this back in 2017. We should let this continue as is and this is applicable for Docker Only environment not for orchestration using K8S. The K8S handles this as part of it's automation.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch
mriccell: Regardless of providing instructions in our K8s projects with instructions and calling it our prefered method of patching it is a requirement that all our images created with Dockerfiles in oracle/docker-imagtes must have a sample which provides instructions how to patch manually in GitHub using Dockerfiles.

OracleSOASuite/dockerfiles/README.md
mriccell: Now you will have different versions 12.2.1.3 and 12.2.1.4, the README in the OracleSOASuite level should be very generic and just indicate that under dockerfiles we have the Dockerfiles and scripts to build binary images of SOA and OSB and that under samples there are instructions and Dockerfiles to patch these GitHub images.  The details of how to build and run should be in the individual 12.2.1.3 and 12.2.1.4 sub directories.

OracleSOASuite/dockerfiles/12.2.1.3/README.md
mriccell: This clarification should be in the README.  I had the question about how to create a OSB domain.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch
mriccell: All our projects should have similar organization otherwise customers fo to the FMW Infra it is one thing and then SOA and it is another.

OracleSOASuite/dockerfiles/README.md
rdas76: For upper stack we're taking a slightly different patching process than Weblogic. In Weblogic for patch image creation your base image is Weblogic released base image and the you're applying the patch.

Refer to line 23 and 40 
https://github.com/oracle/docker-images/blob/master/OracleWebLogic/samples/12213-patch/Dockerfile

For Upper stack our base image if FMW Infra not the Upper stack released image. We're recreating the image with patch applied. This we did based on imagetool create option and your recommendation that updating is going to add to the size.

Refer to line 17 and 54 

 https://github.com/oracle/docker-images/blob/master/OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch

So, for every release we will have 2 sample Docker files, one without patch and one with patch. Hope this clarifies.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch
rdas76: As I mentioned in my other update, Docker image creation on patches we followed the following best practices based on your recommendation -

1) Create the image instead of update. The patch sample that is present in WLS is actually updating the release image with patches. The patch sample that we have written will recreate the image with installers and patches

2) The above create option helps upper stack products to optimize the image sizes otherwise it increases to the size of 10GB or more.

Agree all projects should have similar organization but we need to do something differently as Upper stack products are more bigger in size and complicated. 

Let me know if we can have an offline discussions.

OracleSOASuite/dockerfiles/README.md
mriccell: Ok I agree with this approach but if this is the case then you need to document this in the README.

OracleSOASuite/dockerfiles/12.2.1.3/Dockerfile.patch
mriccell: I agree with this approach.  Is it documented in the README how to patch?

sbattagi: Yes, image creation with patch is documented in the README [link](https://github.com/sbattagi/docker-images/blob/master/OracleSOASuite/dockerfiles/README.md#building-docker-image-with-bundle-patches-for-soa).

rdas76: On further internal discussions, we plan to make Dockerfile.patch as the only Dockerfile. This way we can have only one file and if there are any patches to be applied the customers will need to copy the patches into the predefined location of 12.2.1.3/patches. If any zip file present the image will be created with patches in it. If the folder is empty then image will be created with any patches. Pls review and let us know if this sounds okay.

Djelibeybi: That sounds fine to me, btw.

mriccell: I am fine with this approach as well, please make sure that it is documented in the README.md and you provide an example.

mriccell: I agree with this approach

sbattagi: Checked in the changes for this agreed approach (single Dockerfile for image build with and without patches). Updated the documentation as well.

Djelibeybi:
Your submission has merge conflicts with master. You will need to rebase your branch before we can merge.

sbattagi:
> 
> 
> Your submission has merge conflicts with master. You will need to rebase your branch before we can merge.

Could you please help with the steps for this ?

sbattagi:
> 
> 
> Your submission has merge conflicts with master. You will need to rebase your branch before we can merge.

Could you please help with the steps for this ?

Djelibeybi:
> Could you please help with the steps for this ?

Here's a tutorial from Digital Ocean that steps you through the process: https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request

If this is too complex, I can resolve the merge conflicts myself (after @mriccell approves the code) but it's always better if the folks who own the code do the merge resolution because I may not pick the right side of the conflict.

sbattagi:
> 
> 
> > Could you please help with the steps for this ?
> 
> Here's a tutorial from Digital Ocean that steps you through the process: https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request
> 
> If this is too complex, I can resolve the merge conflicts myself (after @mriccell approves the code) but it's always better if the folks who own the code do the merge resolution because I may not pick the right side of the conflict.

Thanks Avi for sharing the doc. I have now completed the rebase (by manually resolving the conflicts). Please check if things are fine.

Thanks
Samba

Djelibeybi:
Looks great, nice job. It still needs @mriccell to approve the changes she requested before I can perform the merge, though.

PR #1514, opened 2020-01-31
Minor improvements to the shell scripts for oracle database configs

Ma27:
* Use `/usr/bin/env bash` to avoid issues on distros that don't provide
  `/bin/bash` (e.g. NixOS).
* Give a hint how to increase storage-size in dockerd
  if the space-check fails.

### DIFF
--- a/OracleDatabase/RAC/OracleConnectionManager/dockerfiles/12.2.0.1/checkSpace.sh
+++ b/OracleDatabase/RAC/OracleConnectionManager/dockerfiles/12.2.0.1/checkSpace.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # LICENSE UPL 1.0
 #
 # Copyright (c) 1982-2018 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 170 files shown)


Djelibeybi:
I don't understand why you're changing the scripts that run inside the containers: we know what's inside the containers and it will always contain `/bin/bash` so this change is unnecessary.

Ma27:
My mistake - I did a `sed`, although this was only relevant for `buildDockerImage.sh`, sorry!
Will fix that tomorrow :)

Djelibeybi:
I'm going to close this pull request because it has 170 modified files and you probably only need to modify a handful. Likewise, I think that storage test should also be done in `buildDockerImage.sh` because you can determine storage driver externally and warn accordingly.

Djelibeybi:
Please do submit a new PR with an updated commit, though. And in the new PR, can you also confirm whether you've signed the Oracle Contributor Agrement at https://www.oracle.com/technetwork/community/oca-486395.html -- if not, please do sign and submit that before opening the new PR as outlined in the contribution guide for this repo.

PR #1519, opened 2020-02-10
Updated dockertools to container-scripts to be consistent with other products

sbattagi:
Hi,

As suggested by Monica, to be consistent across WLS, FMW Infra and FMW, standardized container-scripts inside SOA image as 'container-scripts' instead of 'dockertools'. 

Thanks
Samba


Djelibeybi:
Please remember to update your master branch from the upstream master before you submit another commit. You should be using feature branches on your clone and keeping your master up to date so that rebasing (when necessary) is simpler.

PR #1521, opened 2020-02-11
fix logic of the script

felix-egli:
2 would never have been returned, because $status would not contain the
value which indicates healty, if $ret is not 0.


Djelibeybi:
Thanks for your contribution. Can you confirm whether you've signed the Oracle Contributor Agreement: https://www.oracle.com/technetwork/community/oca-486395.html? If not, you'll need  review and submit a signed OCA before this pull request can be considered.

Djelibeybi:
Closing due to lack of signed OCA.

Issue #1523, opened 2020-02-12
Coherence 12.2.1.3.2 docker image does not work with OpenShift


dhirupandey:
The current docker image available for Coherence 12.2.1.3.2 does not work with OpenShift platform. 

Similar problems have already been addressed in the recent update to Coherence 12.2.1.4.0 image by #1491 that fixed the OpenShift issues and reduced the size of the image as well.

dhirupandey:
Fixed with PR https://github.com/oracle/docker-images/pull/1524

PR #1524, opened 2020-02-12
Changes for creating 12.2.1.3.2 docker image that will work with OpenShift

dhirupandey:
This fix will address https://github.com/oracle/docker-images/issues/1523

The current scheme of creating 12.2.1.3.2 image by using Coherence 12.2.1.3.0 base image and then patching it with Coherence cumulative patch, bloats the size of the image. 

Creation is now handled by Dockerfiles specialized for this in a separate 12.2.1.3.2 directory and include fixes for running on OpenShift.


Djelibeybi:
You seem to be pushing more commits to your branch. Can you please let us know when this is ready for review?

dhirupandey:
Commits are done now. No longer WIP

PR #1525, opened 2020-02-13
aligning version numbers in script with version directories

r2datical:
buildDockerImages.sh did not align with the directories in the repos.


r2datical:
This was premature. I'm going to resubmit another one. Please kill this.

PR #1537, opened 2020-02-28
Create new db on container start if oradata folder was manually emptied

### LINKED ISSUE #1533: DB does not start any more if oradata folder is emptied with stopped container
I am trying to integrate Oracle Xe in a docker-compose project.

For persisting data, I use a host mount instead of volume for the oradata folder, and it works fine on first start of the container.
However, if I: stop the container, remove all files in the oradata folder, and start it again, the db does not start any more.

The main problem seems to be with how the db is not natively capable of (re)creating the datafiles if those are missing at boot, and what runOracle.sh currently does.

This could be fixable f.e. by keeping copies of empty data files in a separate location and having a shell script running at container start that copies them over if oradata dir is empty. Not that different from what runOracle.sh currently does (except for the fact that runOracle.sh moves the existing files, which makes it work only once and break if the volume gets manually cleaned up before a container restart).
Maybe there is even a better, native solution that I am not aware of that would create data files... my oracle-fu is getting rusty

gggeek:
This fixes #1533 

I have tested the patch only with oracle 18.04 xe, thus the PR only touches the container for that version.

The fix relies on `/etc/init.d/oracle-xe-18c configure` being able to correctly re-create the database from scratch after the files have been removed. I guess that that should work in other db versions too, but there might have been changes to the oracle init script, netca or dbca that I am not aware of...

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/18.4.0/runOracle.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/18.4.0/runOracle.sh
@@ -71,6 +71,27 @@ function symLinkFiles {
    cp $ORACLE_BASE/oradata/dbconfig/$ORACLE_SID/oratab /etc/oratab
 }
 
+########### Opposite of the above ############
+function undoSymLinkFiles {
+
+   if [ -L $ORACLE_HOME/dbs/spfile$ORACLE_SID.ora ]; then
+      rm $ORACLE_HOME/dbs/spfile$ORACLE_SID.ora
+   fi;
+
+   if [ -L $ORACLE_HOME/dbs/orapw$ORACLE_SID ]; then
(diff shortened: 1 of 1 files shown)


Djelibeybi:
@gvenzl we have a signed and processed OCA for this submission. Please go ahead and review.

Djelibeybi:
@gvenzl this pull request has been waiting for your review for _months_. Please either review or reassign.

rishabh20 (approved):
Looks good to be merged for the 18.4 release. Not needed for non-XE releases such as 18.3, 19.3.

yunus-qureshi (approved):
@gvenzl this is good for merge

gvenzl (approved):

gvenzl:
Thanks a lot for your contribution, @gggeek

PR #1548, opened 2020-03-12
Paths in documentation don't match directory structure

sparsick:
I try out readme for prebuildtbd and the described paths don't match the directory structure. I fix it.


gvenzl:
Thank you very much and well spotted, @sparsick!

@Djelibeybi, this is just a minor doc update and shouldn't require an OCA.

sparsick:
Great thanks for your fast accepting :+1:

PR #1549, opened 2020-03-12
Adding Dockerfiles for 20.0.0 release

mzachh:
The 20.x dockerfiles are currently a copy of GraalVM/CE/19 with a different GRAALVM_VERSION var.


Djelibeybi:
If the Dockerfiles are the same, why do we need a different set? Why can't we just set the Graal version at build?

mzachh:
I am unsure how to structure it best, we could also drop the directory for the version in "GraalVM/CE/19".

I slightly prefer having one directory per major version, because 19.3 is a LTS version. So there will be also new 19.x releases this year.
It gives us more flexibility to develop the Dockerfiles for 19 and 20 differently.

Djelibeybi:
I definitely prefer simplicity and deduplication, so if the Dockerfiles are identical except for 19 vs 20, then I'm strongly more in favour of just removing the version from the path completely.

Djelibeybi:
You've changed the files, but not the path? I would suggest removing the `19` from the path completely.

PR #1550, opened 2020-03-12
Removes oracle-database-xe.rpm from Docker Image

andregmoeller:
If I build an Oracle XE Docker Image by calling
```
% ./buildDockerImage.sh -v 18.4.0 -x
```
and then run the image in interactive mode
```
% docker run -it oracle/database:18.4.0-xe /bin/bash
```
I will see that the rather large `oracle-database-xe-18c-1.0-1.x86_64.rpm` is saved in the container:
```
# ls -lah /var/tmp/yum-root-6gKIob/
total 2.4G
drwx------ 2 root root 4.0K Mar 12 12:15 .
drwxrwxrwt 1 root root 4.0K Mar 12 12:35 ..
-rw-r--r-- 1 root root 2.4G Jan 21 15:53 oracle-database-xe-18c-1.0-1.x86_64.rpm
```
This PR deletes the folder `/var/tmp/yum-*`.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/18.4.0/Dockerfile.xe
+++ b/OracleDatabase/SingleInstance/dockerfiles/18.4.0/Dockerfile.xe
@@ -53,6 +53,7 @@ RUN chmod ug+x $INSTALL_DIR/*.sh && \
     yum -y install openssl oracle-database-preinstall-18c && \
     yum -y localinstall $INSTALL_FILE_1 && \
     rm -rf /var/cache/yum && \
+    rm -rf /var/tmp/yum-* && \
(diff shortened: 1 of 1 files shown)


Djelibeybi:
Good catch.

@gvenzl there is no material IP in this change so it should OK to review without an OCA.

gvenzl:
Very good catch indeed!

@Djelibeybi, should we rethink our `yum` cleanup methods?
Any better way to make sure everything is clean after `yum` was invoked?

Djelibeybi:
This is a side-effect of using the `localinstall` mechanism instead of having the RPM in a yum repo. I think this fix is sufficient until the XE RPM is published on yum.oracle.com

PR #1569, opened 2020-04-03
Support SSL in 12213-domain-home-in-image

qvdang:
Test results:
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1235
https://build.weblogick8s.org:8443/job/weblogic-kubernetes-operator-quicktest/1238 (re-run of the 1 test failure in 1235).


qvdang:
Hi @mriccell , please review this PR when you have some time. Thanks!

OracleWebLogic/samples/12213-domain-home-in-image/Dockerfile
```diff
@@ -1,6 +1,6 @@
-#Copyright (c) 2014-2018 Oracle and/or its affiliates. All rights reserved.
+#Copyright (c) 2014-2020 Oracle and/or its affiliates. All rights reserved.
```
rjeberhard, line 1: Format is "2014, 2020," -- it's a comma + space between 2014 and 2020 and then a comma after.

qvdang: I see "2014-2020" in a few other files too.  Let me update those.

Djelibeybi: I hate to be that guy, but there is no comma after and we no longer use "All rights reserved either.". The correct format now is `Copyright (c) 2014, 2020 Oracle and/or its affiliates.`

qvdang: @Djelibeybi , do you have the link to Oracle's official guideline on the copyright statement?

Djelibeybi: Yes, but it's an internal link, so please email me directly and I'll send it to you.

OracleWebLogic/samples/12213-domain-home-in-image/README.md
```diff
@@ -73,6 +76,8 @@ To start the containerized Administration Server, run:
           -v <HOST DIRECTORY TO PROPERTIES FILE>/properties/docker-run:/u01/oracle/properties \
           12213-domain-home-in-image
 
+If both the JAVA_OPTIONS environment variable and the JAVA_OPTIONS in the docker-run/security.properties file are configured, the latter gets appended to the former. If SSL is enabled, pass "-e JAVA_OPTIONS=-Dweblogic.security.SSL.ignoreHostnameVerification=true" to the "docker run ..." command that starts a managed server.  This is necessary because the Demo identity certificate gets generated with a host name different from the admin server host when the domain is created.
```
mriccell, line 79: Could you provide an additional example how to run a Docker command to start the Admin Server and Managed Server wit SSL configured?  This will make it clearer.

qvdang: Monica, this is a domain home in image sample, so SSL is configured(if SSL_ENABLED=true in properties/docker_build/domain.properties) as part of the domain generation in the build step above.  In other words, we don't enable SSL in the docker run command(although we technically can by passing -Dweblogic.ssl.Enabled=true but that's not the point of this sample update).

mriccell: Could you add that description in the README?  Something like:
To enable SSL set `SSL_ENABLED=true` in the properties file  `properties/docker_build/domain.properties`. 
I do not understand from this paragraph why I would want to set "-e JAVA_OPTIONS=-Dweblogic.security.SSL.ignoreHostnameVerification=true" could you clarify?

qvdang: Yes, that has to do with the host name verification as part of the SSL handshake.  If SSL is enabled, the managed server connects to the admin server over SSL(https specifically) to download the domain config when booting up.  Because the demo identity certificate gets generated when the domain is created, the host name of the environment is used as the host name for the certificate(that's the default behavior of WebLogic's utils.CertGen utility).
With -Dweblogic.security.SSL.ignoreHostnameVerification=true, host name verification is bypassed to allow a successful SSL connection.  Hope that's clear enough :).

mriccell: How about replacing the JAVA_OPTIONS paragraph for something like:

 When SSL is enabled and at the time the managed server is coming up it connects to the admin server (over SSL) to download the domain configuration. Because the demo identity certificate gets generated when the domain is created, the host name of the environment is used as the host name for the certificateDemo.   Managed server identity certificate are generated with a host name different from that of the admin server host. To bypass host name verification and allow a successful SSL connection the system property -Dweblogic.security.SSL.ignoreHostnameVerification=true must be set in the "docker run ..." command which starts the managed server. Note that if both the JAVA_OPTIONS environment variable and the JAVA_OPTIONS in the docker-run/security.properties file are configured, the latter gets appended to the former.

Issue #1575, opened 2020-04-14
Make the memory allocated to the database configurable


steveswinsburg:
Currently it is hardcoded to 2048 (unless nproc > 8). The PR for this makes this configurable from the start, but leaving the default at 2048.

gvenzl:
Changes with the memory parameters have been merged.

PR #1576, opened 2020-04-14
Allow the initial amount of memory for databases to be configurable

steveswinsburg:
This is the PR for #1575. 

I have only added this for Oracle 19c. Let me know if this needs to be added for the other versions as well.

Also note I submitted the OCA on 13th April 2020.

### DIFF
--- a/OracleDatabase/SingleInstance/README.md
+++ b/OracleDatabase/SingleInstance/README.md
@@ -59,6 +59,7 @@ To run your Oracle Database Docker image use the **docker run** command as follo
 	-e ORACLE_SID=<your SID> \
 	-e ORACLE_PDB=<your PDB name> \
 	-e ORACLE_PWD=<your database passwords> \
+	-e ORACLE_MEM=<amount of memory to allocate> \
 	-e ORACLE_CHARACTERSET=<your character set> \
 	-v [<host mount point>:]/opt/oracle/oradata \
 	oracle/database:19.3.0-ee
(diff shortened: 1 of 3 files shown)


Djelibeybi:
@gvenzl OCA verified. Please go ahead and review.

gvenzl:
@yunus-qureshi @latakannan @rishabh20oracle, can you please take a look at this PR please and evaluate the ability to pass on memory requirements to the database? Ideally, this parameter takes effect also when a database is started, not only when it's created. The question is whether we can easily retrieve the configured parameter without having to startup the database, retrieve the memory, and then restart the DB with the new value if it has changed.

Also, please have a look at https://stackoverflow.com/questions/42187085/check-mem-limit-within-a-docker-container. We need to also verify whether the container does even have enough memory that is requested by the parameter.

rishabh20:
@gvenzl @yunus-qureshi @latakannan I see a few concerns with this PR and I have few use-cases in mind where this will not address the problems that we intend to solve with this enhancement.

1. The parameter being introduced will only be taken into account if the no. of cores is less than 8. Otherwise, 40% of available memory will be used. And till 19c, the logic in DBCA for getting available memory inside a docker container hasn't been enhanced to understand things like pod yaml specs, docker --memory param, etc. So, for large systems where memory limits are set on docker containers, dbca will be creating db using wrong memory values. This brings me to my 2nd point.

2. Should we use this parameter irrespective of whatever the number of cores is? Whenever this param is set, we try to check if the container has this much available memory and continue/fail accordingly. If it is not set, we do what we were doing till now, use 2048 for cores < 8 and 40% of available memory for cores > 8 (here the memory calculation by dbca could go wrong if mem limits are set for the container and then we request the user to use the exposed memory param).

3. Regarding starting the database, instead of creating one, since we make use of spfile for starting the db which is a binary file and not a text file, we will have to start the db, run the show init param sql command and then compare this value with the newly passed value. We will have to run alter system set init param sql command and then bounce the db to reflect the changes. I don't know if we want to do all this or just allow the user to select the memory only for the 1st time while creating the db. Of course, that limits the user to reuse existing datafiles on the same system and does not let them transfer the files to another machine.

rishabh20:
Closing this PR as the recommended way for giving control to the user to set memory limits is to allow them to set SGA and PGA size. Opened another PR #1742 which implements that functionality.

PR #1665, opened 2020-07-22
Remove instruction about downloaded file location

jahson:


Djelibeybi:
Thanks for spotting this.

Issue #1669, opened 2020-07-29
Creating just one single docker image and allowing the user to select the oracle edition (EE/SE) during runtime


rishabh20oracle:
Starting from 12.2.0.1, the binaries for enterprise and standard editions have been the same. Still, separate docker images are released for enterprise and standard editions. A runtime argument can be provided to the user to select the edition using a single docker image.

gvenzl:
Merged.

PR #1670, opened 2020-07-29
Exposing an env variable ORACLE_EDITION to the user to select the edition of the db while running the image

rishabh20oracle:


rishabh20oracle:
Opened issue #1669 for this PR.

rishabh20oracle:
Added a new script called relinkOracleBinary.sh which is copied under oracle base during docker image build and called during runtime just before the database is created/started. This script calls the make command to relink the binaries in accordance with the edition passed by the user while running the image.

gvenzl:
Hey @rishabh20oracle, we need to modify the `buildDockerImage.sh` script to reflect that change as well.
For the versions where editions can dynamically be relinked, we should no longer add `-<edition>` to the end of the tag.
I.e. `oracle/database:19.3.0-ee` --> `oracle/database:19.3.0`

I recommend that we:

1. [Set enterprise edition as the default option](https://github.com/oracle/docker-images/blob/master/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh#L144-L145)  
This should also produce an output informing the user that enterprise edition has been set as default as nothing was provided
1. [Not add the `-$EDITION` for the versions where this change is affected](https://github.com/oracle/docker-images/blob/master/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh#L168-L169)
1. [Amend the usage help to reflect the default](https://github.com/oracle/docker-images/blob/master/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh#L21)
1. Also add this modification to 12.2.0.1 given that this version also already supports this and the changes are limited.

@Djelibeybi, do you have any concerns with relinking binaries during the container runtime? See [here](https://github.com/oracle/docker-images/pull/1670/files#diff-40e73e049e5ae6c3770ff4d5aaf796abR28)

Djelibeybi:
> Hey @rishabh20oracle, we need to modify the `buildDockerImage.sh` script to reflect that change as well.
> @Djelibeybi, do you have any concerns with relinking binaries during the container runtime? See [here](https://github.com/oracle/docker-images/pull/1670/files#diff-40e73e049e5ae6c3770ff4d5aaf796abR28)

No, I don't. I think this is a pretty nifty option to allow a single image to be used for either `se` or `ee` purposes.

rishabh20oracle:
The changes we have made here support choosing the edition during runtime. We have not removed the logic wherein during build time, the edition is first replaced in the db_inst.rsp file which is used to install the software. So the user can still pass the -s option with buildDockerImage.sh and build an image with standard edition software installed. Therefore, we should not remove the $EDITION part from the image name, and cannot/should not set enterprise as the default option.

rishabh20oracle:
Regarding extending these changes to 12.2.0.1 and 18c, QE has only tested these changes for 19.3. We do not know if there is anything else that we would need to address for the earlier releases. Maybe we can do this later as a separate PR once we get this tested for the releases in question.

yunus-qureshi:
> The changes we have made here support choosing the edition during runtime. We have not removed the logic wherein during build time, the edition is first replaced in the db_inst.rsp file which is used to install the software. So the user can still pass the -s option with buildDockerImage.sh and build an image with standard edition software installed. Therefore, we should not remove the $EDITION part from the image name, and cannot/should not set enterprise as the default option.

To clarify the runtime edition switch is a new container runtime option that users may need to get familiar with. The current options to build edition specific images can remain as the existing releases have been out there for sometime and users/scripts are already consuming it. In next DB release we can drop the build time edition support totally.

gvenzl:
So a user can spin up a database container with SE2 from an EE image and vice versa. What's the point of this runtime switch then? Now the user sees a container running from an EE image that's not necessarily EE any longer. How does the user figure out what edition is running? How do we make sure that if a new container is spun up again, that the edition still matches with the database datafiles?

rishabh20oracle:
Q: How does the user figure out what edition is running?
A: I have updated the readme mentioning that the user can see the edition in the database logs just like we print the randomly generated password.

Q: How do we make sure that if a new container is spun up again, that the edition still matches with the database datafiles?
A: I have updated the relinkOracleBinaries.sh script to verify that the oracle edition being passed or the edition of the software home (if oracle edition is not passed) matches with the datafiles. If not, we will error out and prompt the user to pass the correct ORACLE_EDITION.

yunus-qureshi:
> So a user can spin up a database container with SE2 from an EE image and vice versa. What's the point of this runtime switch then? Now the user sees a container running from an EE image that's not necessarily EE any longer. How does the user figure out what edition is running? How do we make sure that if a new container is spun up again, that the edition still matches with the database datafiles?

@gvenzl this runtime edition option serves the purpose of publishing a single image in container registry. Users can still build edition specific images from existing build instructions and use those without worrying about this edition option. And can still choose to change edition on first run instead of rebuilding. This switch is only really useful when DB is created on first run. It can't really switch the edition on existing datafiles. We have added code to error out in such cases if an incorrect edition is passed.

gvenzl:
Ok thanks, we needed that additional check to make sure users cannot shoot themselves in the foot. But still, what's the rationale of having a single image on the container-registry and not the EE and SE2 versions?
While I find the single image cool and we should definitely offer it, what would be the default edition for that?
And how long does the relinking take?
The last thing we want is using not having to waste time building an image and then having to waste time relinking the binaries.

rishabh20oracle:
> Ok thanks, we needed that additional check to make sure users cannot shoot themselves in the foot. But still, what's the rationale of having a single image on the container-registry and not the EE and SE2 versions?
> While I find the single image cool and we should definitely offer it, what would be the default edition for that?
> And how long does the relinking take?
> The last thing we want is using not having to waste time building an image and then having to waste time relinking the binaries.

Yeah, the additional check was much needed. The rationale behind having a single image is that ideally, that is what we should have been doing since the software has no difference between EE and SE since 12.2.0.1. And since 19.3 scripts were already there on GitHub for a long time, we didn't remove the option of building the standard image as users must have been using it. From 20c, we can choose to stop giving the option of building both SE and EE and just keep EE as default. Till then, the users can get used to using the ORACLE_EDITION parameter.

As far as releases on container-registry are concerned, we'll just upload the EE image and the documentation for SE will point to using the EE image with the ORACLE_EDITION parameter. The relinking takes around 20sec max., so yes we did keep in mind that the user doesn't waste time, while thinking of introducing this feature.

gvenzl:
Great, thank you @yunus-qureshi. I recommend that we put all three images on the container registry for 19c then, one without any edition, one `-ee` and one `-se2`.
Rational being the same as you said above, users have gotten used to `-ee` and `-se2` and for long wanted these images provided by a registry, there is no downside on hosting three images for 19c and make the transition smooth for users.
All we waste is a bit more storage on the container registry but for the benefit of staying consistent with the build scripts and the automation scripts that users have already.

For future versions, we can then take a new approach as you say.

latakannan:
Gerald

Given that we have tested just the EE version completely, I was hoping 
we could upload the EE version in both Enterprise and Standard buckets 
in OCR.  We could call out the flag in the README for the SE that could 
be used to bring the DB up in SE mode. The EE edition image is what has 
been extensively tested by QE and the flag to bring up SE has also been 
tested by them.

As far as the consumers are concerned, they should ideally not see too 
much of a difference, since they would see an image in each bucket 
(EE/SE), just that the image tags would be the same in both.  And the 
README for SE would be a little different in terms of how one brings up 
the SE edition using the same image.

Do you see any issues with this approach?

Thanks
--lata

latakannan:
Including all our official ids..

Gerald

Given that we have to upload by next week, we have tested the EE image 
fairly thoroughly and are a little more confident in uploading this at 
this point in OCR. So, it might be safer to just upload this against 
both EE and SE2 and then update the readme for SE2.  If, there indeed is 
a need for a separate image for SE2, then we might not be able to do 
this by Aug 13th. We could upload only the EE version for Aug 13 and do 
the SE2 later.

Thanks
--lata

PR #1683, opened 2020-08-13
Switch WLS installers to lite installers

mriccell:
Switch WLS installers in 12.2.1.4 and 14.1.1.0 to lite installers.  Change permissions of the DOMAIN_HOME.


mriccell:
@ddsharpe can you please review this change?

PR #1701, opened 2020-09-03
Changes to account for new Go 1.15 release on yum.oracle.com

totalamateurhour:


Djelibeybi (approved):
Thanks @totalamateurhour

Issue #1736, opened 2020-10-14
Archiving old content into a non-root location


Djelibeybi:
/cc @gvenzl @mriccell 

I am going to move the olded/abandoned content out of the root of the repo and into an `archive` directory. Initial targets are:

* `ContainerCloud`
* `OracleDataIntegrator`
* `OracleEDQ`
* `OracleTuxedo`

I will open a PR with this change as well.

PR #1737, opened 2020-10-14
Move old content into an Archive location and update the docs.

Djelibeybi:
This implements my recommendation in #1736 by moving old content into an archive location. Added some docs for this and updated `CODEOWNERS` to remove references to archive content so the admins are flagged for future issues for that section of code.


gvenzl (approved):

toddinpal:
Why was Tuxedo moved to archive?  The images that are here are the latest images.

PR #1743, opened 2020-10-16
Instant Client comment and doc updates

cjbj:
"Docker" -> "Container" changes, as hinted at on Slack.

### DIFF
--- a/OracleInstantClient/README.md
+++ b/OracleInstantClient/README.md
@@ -1,6 +1,6 @@
-# About this Docker Image
+# About this Container Image
 
-These Docker images contain the Oracle Instant Client 'Basic', 'SDK' and 'SQL\*Plus' packages.  They can be used to build and run Oracle Call Interface (OCI), Oracle C++ Call Interface (OCCI), and JDBC-OCI applications applications.  The SQL\*Plus command-line query tool allows quick ad-hoc SQL and PL/SQL execution.  The Docker images can be extended with optional packages for ODBC, or to include tools such as Oracle SQL\*Loader.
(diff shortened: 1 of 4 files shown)


OracleInstantClient/README.md
```diff
@@ -1,6 +1,6 @@
-# About this Docker Image
+# About this Container Image
```
Djelibeybi, line 1: Neither container nor image are proper nouns so neither of them should be capitalised.

OracleInstantClient/dockerfiles/12.2.0.1/Dockerfile
```diff
@@ -1,11 +1,9 @@
 # LICENSE UPL 1.0
 #
-# Copyright (c) 2014-2015 Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2014, 2020 Oracle and/or its affiliates. All rights reserved.
```
Djelibeybi, line 3: If you're changing copyright notices anyway, please remove the "All rights reserved." phrase as well. We no longer include that phrase.

OracleInstantClient/dockerfiles/18/Dockerfile
```diff
@@ -1,11 +1,8 @@
 # LICENSE UPL 1.0
 #
-# Copyright (c) 2014, 2019, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2014, 2020, Oracle and/or its affiliates. All rights reserved.
```
Djelibeybi, line 3: If you're changing copyright notices anyway, please remove the "All rights reserved." phrase as well. We no longer include that phrase.

Djelibeybi (changes requested):
Considering this is purely a documentation update, I went into full copy editor mode. :)

OracleInstantClient/README.md
cjbj: You're supposed to be away aren't you? It's a heading IMHO.  Hence the '#'

Djelibeybi: Oracle standard is sentence case for headings which means only the first word is capitalised.

cjbj: See Slack

Djelibeybi (approved):

PR #1744, opened 2020-10-16
update COH sample Dockerfile to match base image oracle:root

ddsharpe:
Base image was changed to oracle:root, so this sample Dockerfile needed to be updated to match.


Djelibeybi (approved):

Djelibeybi:
@mriccell can you please review?

PR #1785, opened 2020-11-11
Changed shebang line to #!/usr/bin/env bash

phdoerfler:
This allows this script to run on systems where bash is not installed in /bin/bash such as NixOS

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
@@ -1,4 +1,5 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+set -e
 # 
 # Since: April, 2016
 # Author: gerald.venzl@oracle.com


Djelibeybi:
If we're going to change one `buildDockerImage.sh` then we should change _all_ of them for consistency. Also, using the script isn't mandatory and we only test it on Oracle Linux, so I'm not sure this is a valid concern.

Could you please open an issue for this as well, so we can discuss whether this is the right solution to provide build scripts for non-OL platforms?

Djelibeybi:

phdoerfler:
Sorry for the late response, life's been very busy recently.
I was unaware that this was only tested and by implication "supported" on Oracle Linux. Had I known this I wouldn't have bothered with this PR in the first place. It's only logical to thus close this issue, indeed.

If it weren't for that, the argument could be made that unless it is important that a specific version of `bash` was used **and** that it is also important that the script would fail if this version of `bash` was **not** living in `/bin/bash`, using `env` would be the better choice: Not having the aforementioned requirements – and I do not see these reflected in the actual script – grants the freedom of using *"any"* (actually the first) `bash` in `$PATH`, which is precisely what `env bash` is doing. Not only does this increase portability, which is apparently not a concern for this repository, it also serves a different purpose:
It indicates to a developer that the exact whereabouts of the `bash` interpreter are irrelevant. In contrast `/bin/bash` implies that it is crucial for the script that `bash` is in that exact location. It is unnecessarily specific and, if you want, misleading.

That's just my two cents. Also: happy holidays :)

PR #1811, opened 2020-12-16
updated base image from 7-slim to 8-slim,

aureliogrb:
changed yum to microndf, and removed jdk/openjdk 14 dockerfiles as that release has reached eol


Djelibeybi:
You've just submitted the same PR that I rejected before. You cannot change the existing Dockerfiles to use `oraclelinux:8-slim` as this will break all downstream customer images that expect a particular base OS version. You must _add_ new Dockerfiles for the Oracle Linux 8 based variant to give customers enough time to switch.

PR #1825, opened 2021-01-05
Improve podman detection for the OracleDatabase/SingleInstance buildDockerImage.sh  script

AmedeeBulle:
The `buildDockerImage.sh` script does not detect properly Podman on OL8 as:
- `docker version --format '{{.Server.Version | printf "%.5s" }}'` does return a version (OL7 version does not)
- `docker info --format '{{.host.BuildahVersion}}'` is an invalid format (should be `{{.Host.BuildahVersion}}`)

This PR address both issues for OL8 (podman 2.0.5) while keeping compatibility with OL7 (podman 1.6.4)

--

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/buildDockerImage.sh
@@ -52,7 +52,8 @@ checksumPackages() {
 checkPodmanVersion() {
   # Get Podman version
   echo "Checking Podman version."
-  PODMAN_VERSION=$(docker info --format '{{.host.BuildahVersion}}')
+  PODMAN_VERSION=$(docker info --format '{{.host.BuildahVersion}}' 2>/dev/null ||
+                   docker info --format '{{.Host.BuildahVersion}}')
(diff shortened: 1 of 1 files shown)


gvenzl:
Thanks @AmedeeBulle!

Is there any slick way we can change not to rely on the podman `docker` alias, in case somebody didn't install that and comes straight via `podman`?

AmedeeBulle:
I was thinking about that, but I didn't wanted to make too many changes at first.

Commands are exactly the same -- that is, the alias isn't doing anything else than calling `podman` with the same parameters.

So we could check at the very beginning if we have a `podman` executable if we don't have a `docker` one and put that in a variable (we should do it in that order as you could have both installed...).

I can prepare another PR with that if you want.
(This one is still needed as we don't know if the user has docker **and** podman installed or only podman with the docker alias...)

gvenzl:
Hey @AmedeeBulle,

> I can prepare another PR with that if you want.
(This one is still needed as we don't know if the user has docker and podman installed or only podman with the docker alias...)

That would be amazing, yes please! :)

gvenzl (approved):

PR #1829, opened 2021-01-07
Database: improve support for Podman

AmedeeBulle:
The `buildDockerImage.sh` script will look for both Docker and Podman in the PATH.
If both are available, Docker is used.

Tested on

- OL7 with Docker and Podman
- OL8 with Podman, with and without the `podman-docker` package

--


gvenzl:
This looks great, thanks @AmedeeBulle!

Should we go all the way and start replacing "Docker" with "Container" where it makes sense?
I.e. also rename the file `buildContainerImage.sh`, etc.?

What do you think?

Djelibeybi:
While we're at it, Oracle corporate style has switched fully to sentence case everywhere, so please can we lose all the capital letters? :) Neither container nor image are proper nouns.

gvenzl:
The casing aids readability, nothing more and nothing else.

Djelibeybi:
Except it doesn't and that's what Oracle has specifically mandated that we should not longer be using title case.

AmedeeBulle:
Although replacing _Docker_ by _container_ makes sense, I am not sure how far we should go with that...
The SingleInstance README page has 60+ references to _Docker_. In most case _Docker Container_ can be simply replaced by _container_, but what are we doing with all the _code blocks_?
Add a paragraph to state Podman can be used instead?

@Djelibeybi any recommendations on how to make all this more builder / runtime agnostic?
(and what about the repo name ? ;) )

Djelibeybi:
(The repo name bugs me too, but we already have a `container-images` repo and renaming that one would be tedious).

We don't have to change any code blocks or example commands. Using `docker` to demonstrate or document how to build or run is fine. We should only be removing Docker when used as an adjective.

Having Podman examples would be nice, but given the compatibility aliases, are not mandatory or urgent.

AmedeeBulle:
Script renamed and doc updated taking in account the above remarks.

Added a paragraph in the FAQ about the `uid` remapping for rootless containers.

gvenzl (approved):
Awesome, thanks so much @AmedeeBulle!

PR #1831, opened 2021-01-12
[docs] The org.opencontainers.image labels need to be mandatory

Djelibeybi:
The `oraclelinux:7-slim` and `oraclelinux:8-slim` images are now
shipping with these labels. If a product does not provide more
specific values, then it will appear to products that consume
these labels as "Oracle Linux" and not a more specific product image.

This is being noticed in particular for images that are published to
GitHub Container Registry that does consume these tags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,10 +94,61 @@ the image owner to implement it.
 1. All images must provide a `CMD` or `ENTRYPOINT`. If your image is designed
 to be extended, then this should output documentation on how to extend the
 image to be useful.
-1. Use `LABEL` instructions for additional information such as ports and volumes.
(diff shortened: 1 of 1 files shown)


mark-au:
If I read this correctly, the OL slim images are shipping with these tags therefore we must overwrite them to avoid muddying users metadata.

Djelibeybi:
Correct.

mark-au (approved):
If I read this correctly, the OL slim images are shipping with these tags therefore we must overwrite them to avoid muddying users metadata.

Djelibeybi:
After some internal discussions, it has been decided that use of these labels is not customer-friendly, so instead of mandating them downstream, I will be removing them from the base Oracle Linux images instead.

PR #1836, opened 2021-01-13
[fix] GHCR does not support GITHUB_TOKEN so workarounds are required

Djelibeybi:
Currently we're using a Personal Access Token to push images to
GitHub Container Registry, but we want the images to be pushed to the
right namespace.


AmedeeBulle (approved):
This generates an issue when the repo owner has uppercase letters, but this is probably good enough for now...

PR #1841, opened 2021-01-14
[fix] Remove org.opencontainers.image labels from all Dockefiles

Djelibeybi:
Because LABELs are automatically inherited and cannot be removed
by downstream images, we are removing them from our images instead.
This is to avoid a downstream image misidentifying itself as an
Oracle produced artefact.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,30 +135,6 @@ LABEL "provider"="Oracle"                                   \
       "port.apex"="8080"
 ```
 
-You may also chose to use the [OpenContainer `image-spec`](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys)
-pre-defined annotation keys:
-
-* **org.opencontainers.image.created** date and time on which the image was
-  built (string, date-time as defined by [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)).
-* **org.opencontainers.image.authors** contact details of the people or
-  organization responsible for the image (freeform string)
-* **org.opencontainers.image.url** URL to find more information on the image
-  (string)
-* **org.opencontainers.image.documentation** URL to get documentation on the
-  image (string)
-* **org.opencontainers.image.source** URL to get source code for building the
-  image (should be the URL of this repository).
-* **org.opencontainers.image.version** version of the packaged software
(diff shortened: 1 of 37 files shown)


cjbj (approved):
My bits look OK via visual inspection of the diff.

gvenzl:
It appears that this fix should be done for actually all Dockerfiles that are in this repository if we think that these `LABEL` instructions are a concern and also be rolled out then to the currently existing images on container-registry.oracle.com.

@yunus-qureshi, do we have any feedback on whether customers are relying on the `LABEL`s inside the Oracle Database Docker image and if so, how big the impact would be if we were to remove these?

Djelibeybi:
@cjbj you'll need to review again asn @gvenzl requested another review from you. I'll also need @NoSQLGuy to approve as this touches the `NoSQL` files. 

@gvenzl this PR is complex enough with multiple reviewers. While I agree that `LABELS` are problematic, I think the rest can be done by individual product owners so the approval and merge is simpler.

rjeberhard (approved):

cjbj (approved):
My bits look OK via visual inspection of the diff.

yunus-qureshi:
> It appears that this fix should be done for actually all Dockerfiles that are in this repository if we think that these `LABEL` instructions are a concern and also be rolled out then to the currently existing images on container-registry.oracle.com.
> 
> @yunus-qureshi, do we have any feedback on whether customers are relying on the `LABEL`s inside the Oracle Database Docker image and if so, how big the impact would be if we were to remove these?

@gvenzl, we don't have any such feedback. So I think it won't be a big impact. Are we going to do the same for the Docker files under OracleDatabase/SingleInstance/dockerfiles ? or may be instead of removing the entire LABEL instruction we can just remove the keys "provider" and "issues"

yunus-qureshi (approved):
Docker files under OracleDatabase/SingleInstance/Dockerfiles do not have the org.opencontainers.image labels. But have the label provider=Oracle. Do we need to remove such labels too?

Djelibeybi:
@gvenzl @mriccell @NoSQLGuy  could you please review ASAP?

ddsharpe (approved):

mark-au (approved):

Djelibeybi:
It's been over a week with no feedback, so I'm going to go ahead and merge.

PR #1851, opened 2021-01-27
ant jar path changed to Weblogic's oracle_common

prshshuk:
This transaction consists wcs-wls-docker-install installer (jar) Manifest file changes. The wcs-wls-docker-install.jar is the one that does the RCU and the ConfigWizard step for Oracle WebCenter Sites.
Ant jars need to be referred from Weblogic's oracle_common libraries as it's no longer shipped with Sites product.


Djelibeybi (approved):

Djelibeybi:
In future, please ensure that your `git config` for `user.email` is correct and that you `--signoff` or -`S` your commits.

PR #1867, opened 2021-03-04
Update setPassword.sh : escape quotes in order to protect passwords which start with a number

gfrann:
escape quotes in order to protect passwords which start with a number


Djelibeybi:
@gvenzl this can be reviewed/merged without an OCA. Which I think it should be.

gvenzl (approved):

Issue #1876, opened 2021-03-22
Scripts to support SI DB docker images deployment in kubernetes setup


rishabh20:
We are checking in the extension that packages the scripts needed to support the deployment of oracle single instance database docker images in a Kubernetes cluster.

PR #1877, opened 2021-03-22
Kubernetes support for SI DB docker images 

### LINKED ISSUE #1866: chart not find  checkDBLockStatus.sh
location: docker-images/blob/master/OracleDatabase/SingleInstance/helm-charts/oracle-db/templates/deployment.yaml
checkDBLockStatus.sh not find

rishabh20:
Opened #1876 for this PR.
Fixes #1866


gvenzl:
Is there any reason why we would not just add this to the actual image rather than making it an extension? There seems to be no harm done when running outside a K8s environment, or?

Djelibeybi:
As you are pushing more changes to this PR, I've switched it to draft mode. Once you feel it is ready for review, you can push the handy "Ready for review" button.

gvenzl (dismissed):

yunus-qureshi (approved):

yunus-qureshi:
> Is there any reason why we would not just add this to the actual image rather than making it an extension? There seems to be no harm done when running outside a K8s environment, or?

@gvenzl even running in K8s does not need this extn as long as you keep the deployment to a single replica. This is needed only when you need to run it in multiple replica mode for HA/fail over etc. We may add this to base in future releases. The PR is good for merge

abhisbyk (approved):

gvenzl (approved):

Issue #1882, opened 2021-03-24
[Oracle 18c] Fail to restart


ortizman:
I started the container, created the database and worked on it without problems. 
I even restarted to check that the docker volume was well mapped and that the data was not lost.

The next day, oracle no longer started and gave me the following message

Can you help me with this error?

```
The Oracle base remains unchanged with value /opt/oracle
#####################################
########### E R R O R ###############
DATABASE SETUP WAS NOT SUCCESSFUL!
Please check output for further info!
########### E R R O R ###############
#####################################
The following output is now a tail of the alert.log:
ORA-01157: cannot identify/lock data file 13 - see DBWR trace file
ORA-01110: data file 13: '/opt/oracle/product/18c/dbhomeXE/dbs/TBS_ABC_RELEASE.DBF'
ORA-1157 signalled during: ALTER DATABASE OPEN...
2021-03-24T12:36:47.866000+00:00
Errors in file /opt/oracle/diag/rdbms/xe/XE/trace/XE_mz00_110.trc:
ORA-01110: data file 13: '/opt/oracle/product/18c/dbhomeXE/dbs/TBS_ABC_RELEASE.DBF'
ORA-01565: error in identifying file '/opt/oracle/product/18c/dbhomeXE/dbs/TBS_ABC_RELEASE.DBF'
ORA-27037: unable to obtain file status
Linux-x86_64 Error: 2: No such file or directory
Additional information: 7
```

***startup docker command***
`docker run --rm --name oracle-18xe -p 1521:1521 -p 5500:5500 -e ORACLE_PWD=Prueba123 -e ORACLE_CHARACTERSET=UTF8 -v /home/manuel/.oracle_xe_oradata/:/opt/oracle/oradata oracle/database:18.4.0-xe`

**folder permissions**

_ls -lart_
`drwxrwxrwx   4  54321  54321    4096 mar 10 17:22  .oracle_xe_oradata
`

**Docker Info**

```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)

Server:
 Containers: 4
  Running: 0
  Paused: 0
  Stopped: 4
 Images: 36
 Server Version: 20.10.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
[...]

PR #1885, opened 2021-03-30
Update FAQ for #1882

gvenzl:
Add an FAQ for the question in issue #1882.

### DIFF
--- a/OracleDatabase/SingleInstance/FAQ.md
+++ b/OracleDatabase/SingleInstance/FAQ.md
@@ -74,3 +74,7 @@ echo "DISABLE_OOB=ON" >> ~/.sqlnet.ora
 ```
 
 For more information configuring _sqlnet.ora_ file see [Database Net Services Reference](https://www.oracle.com/pls/topic/lookup?ctx=dblatest&id=GUID-2041545B-58D4-48DC-986F-DCC9D0DEC642), [Instant Client Installation for Linux](https://www.oracle.com/database/technologies/instant-client/linux-x86-64-downloads.html), [What is DISABLE_OOB (Out Of Band Break)? (Doc ID 373475.1)](https://support.oracle.com/epmos/faces/DocumentDisplay?id=373475.1) and issue #1352
+
+## ORA-01157: cannot identify/lock data file
+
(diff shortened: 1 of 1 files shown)


OracleDatabase/SingleInstance/FAQ.md
```diff
@@ -74,3 +74,7 @@ echo "DISABLE_OOB=ON" >> ~/.sqlnet.ora
 ```
 
 For more information configuring _sqlnet.ora_ file see [Database Net Services Reference](https://www.oracle.com/pls/topic/lookup?ctx=dblatest&id=GUID-2041545B-58D4-48DC-986F-DCC9D0DEC642), [Instant Client Installation for Linux](https://www.oracle.com/database/technologies/instant-client/linux-x86-64-downloads.html), [What is DISABLE_OOB (Out Of Band Break)? (Doc ID 373475.1)](https://support.oracle.com/epmos/faces/DocumentDisplay?id=373475.1) and issue #1352
+
+## ORA-01157: cannot identify/lock data file
+
+This error occurs when the database cannot find a data file (used for tablespaces) that was previously present. This is most likely because the data file has been located outside the volume in a previous container and was hence not persisted. Ensure that when you add tablespaces and/or data files that they are located within the volume location, i.e. $ORACLE_BASE/oradata/$ORACLE_SID, (e.g. `/opt/oracle/oradata/XE`).
```
Djelibeybi, line 80: Can you provide an actual sample SQL command that should be used here?

Djelibeybi (changes requested):

OracleDatabase/SingleInstance/FAQ.md
gvenzl: No, I cannot. There are so many parameters with the `CREATE TABLESPACE` clause and it depends on so many factors of what the user wants to do, that I'm hesitant to just pick one that might work that users than just copy/paste. It starts with whether the user wants to create a data, temp, or undo tablespace, whether it's a bigfile tablespace or smallfile one, what the tablespace and file(s) name(s) should be, the size, the autoextend parameters, whether the tablespace should be encrypted, compressed, and about another 100 different parameters. Furthermore, a user could happily just put the data files in a separate volume as well, they don't have to go into that particular directory or any other. Users should be aware that if they create files on the file system in a Docker container, these files have to either remain in a volume or are inherently transient by nature. Given that in 5 years of existence of these scripts this has come up only once and only now, I think it is fair to say that most users know that if they type the `CREATE TABLESPACE` clause, they can and probably should lookup that clause in the documentation of the product, rather than copy/pasting it from a FAQ markdown file for Docker scripts.

Djelibeybi (approved):

PR #1889, opened 2021-03-31
Basic workflow to build Instant Client 19 and 21 GHCR images

cjbj:
A simplistic action that builds Instant Client 19 & 21 images for GHCR


Djelibeybi (approved):
I want @AmedeeBulle to review too, before I merge. Once that's done, and the image is built, I'll get you to do a final sanity check before making the image public.

AmedeeBulle (approved):
LGTM

(`ghcr.io` repo is hard-coded, so you can't really test in a fork, but I guess that's OK, it is not the only workflow doing that)

PR #1895, opened 2021-04-05
To Deploy kubernetes pods of OracleDatabase only on nodes in a particular nodepool having label "pool"

Mahi0911:
To Deploy Kubernetes pods of OracleDatabase only on nodes in a particular NodePool having label "pool" , Added label "pool" under nodeSelector in OracleDatabase/SingleInstance/helm-charts/oracle-db/templates/deployment.yaml

### DIFF
--- a/OracleDatabase/SingleInstance/helm-charts/oracle-db/values.yaml
+++ b/OracleDatabase/SingleInstance/helm-charts/oracle-db/values.yaml
@@ -47,7 +47,7 @@ imagePullPolicy: Always
 imagePullSecrets:
 
 ## Deploy only on nodes having required labels .
-## Format label_name : label_value . eg pool : sidb
+## Format label_name : label_value . eg pool: sidb
 ## Leave empty if there is no such requirement
-nodeLables:
+nodeLabels:
 #  pool: sidb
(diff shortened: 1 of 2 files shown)


yunus-qureshi (commented):
@Mahi0911 can we generalise the label name here instead of assuming to be pool always

gvenzl (changes requested):
As per @yunus-qureshi:
> @Mahi0911 can we generalise the label name here instead of assuming to be pool always?

OracleDatabase/SingleInstance/helm-charts/oracle-db/values.yaml
```diff
@@ -46,3 +46,8 @@ imagePullPolicy: Always
 ## container registry login/password
 imagePullSecrets:
 
+## Deploy only on nodes having required labels .
+## Format label_name : label_value . eg pool : sidb
+## Leave empty if there is no such requirement
+nodeLables:
```
yunus-qureshi, line 52: @Mahi0911 plz fix this typo. nodeLables should be nodeLabels. Update all references

yunus-qureshi (approved):
Looks good for merge

gvenzl (approved):

PR #1903, opened 2021-04-20
ci: update build and push workflows to use GITHUB_TOKEN

Djelibeybi:
Also switched to using `${GITHUB_ACTOR,,}`  in the IC and NoSQL workflows so that builds work for users who fork this repo to their own GitHub account.

I will let @AmedeeBulle review and update the development images workflow.

### DIFF
--- a/.github/workflows/build-and-push-dev-images.yml
+++ b/.github/workflows/build-and-push-dev-images.yml
@@ -146,8 +146,8 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
-          username: ${{ secrets.GHCR_USER }}
-          password: ${{ secrets.GHCR_TOKEN }}
+          username: ${{ github.actor }}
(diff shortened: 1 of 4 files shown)


Djelibeybi:
Once this is merged, I will delete the `GHCR_*` secretys from this repo.

cjbj (approved):

AmedeeBulle:
Have you tested human-triggered builds?

From the [docs](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables):
> GITHUB_ACTOR | The name of the person or app that initiated the workflow. For example, octocat.

My understanding is that if I trigger a build (in the _oracle_ org), I will be the actor and the image will be pushed to my repo, which is not what I want (I use the GitHub API to keep the images current)

Djelibeybi:
> Have you tested human-triggered builds?

Only in my own repo. I'll push to a branch tomorrow and test it from there.

PR #1945, opened 2021-05-24
Remove -i from documented build command

gvenzl:
The `-i` option should not be used in copy/pastable code within the documentation.
The checksum functionality (which `-i` skips) is there to ensure users have downloaded the correct (and not corrupt) Oracle Database ZIP file which is needed to build an image.


yunus-qureshi (approved):
Agreed

PR #1949, opened 2021-06-03
update the name of image serverjre8 in docker.io

jeanfbs:


Djelibeybi:
Oracle no longer controls the `store/oracle` images and do not recommend they are used as base images. The `Dockerfiles` in this repo use the `oracle/` prefix to identify images that should be built locally using another `Dockerfile` _from this repo_. 

In this case, you should build the [`OracleJava/8`](https://github.com/oracle/docker-images/tree/main/OracleJava/8) first as `oracle/serverjre:8` and then build the WebLogic images.

PR #1967, opened 2021-06-10
Fix user script mount issues related to extensions

abhisbyk:
If users mount `/opt/oracle/scripts/setup` and` /opt/oracle/scripts/startup` directories from host machine while creating the container from the extended image (either using patching or k8s extension), some scripts which are already present in those locations (kept there while extending the base docker image) are overwritten.

So created new directories `/opt/oracle/scripts/extensions/setup` and `/opt/oracle/scripts/extensions/startup` to place the scripts while extending the base image. Now, It doesn't get overwritten even if user mount the directories mentioned above.


yunus-qureshi (approved):
Looks fine to me

gvenzl (approved):

PR #1970, opened 2021-06-10
feat(developper): Add oracledb flavor for python 3.8/3.9 on OL8

AmedeeBulle:

### DIFF
--- a/OracleLinuxDevelopers/oraclelinux8/python/3.8-oracledb/Dockerfile
+++ b/OracleLinuxDevelopers/oraclelinux8/python/3.8-oracledb/Dockerfile
@@ -1,0 +1,17 @@
+# Copyright (c) 2020 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+FROM oraclelinux:8-slim
+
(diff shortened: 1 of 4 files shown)


Djelibeybi:
Sorry, but our approval to distribute these images is based on only installing software as RPMs from yum.oracle.com. We are not allowed to use `pip` to install packages.

AmedeeBulle:
Yes, I was thinking about that as well.

Most python developers will have cx_oracle as dependency in their `requirement.txt` or `setup.py`  together with their other dependencies.

So it might make sense to publish with the IC, but without CxO.

But we need to be clear on what we publish to avoid confusion between the the projects with or without CxO...

Djelibeybi:
We already have an Instant Client image that folks can add any language they like on top of. I do not want to publish language-specific images that do not include the language specific driver for Oracle DB.

cjbj:
@Djelibeybi you already have a [lot of language specific images](https://github.com/oracle/docker-images/tree/main/OracleLinuxDevelopers) that don't have the Oracle DB driver, so I assume you mean "language specific, with Instant Client, but without the DB driver".

Djelibeybi:
Yes, that's what I meant. Thanks @cjbj

Issue #1969, opened 2021-06-10
DB crash when container is killed/stopped before DB is ready .


Mahi0911:
If one of the database pods which is initialising database , is killed/stopped before its ready , the other pods try to create the database and result in Pod Crash .

PR #1971, opened 2021-06-11
DB crash when container is killed/stopped before DB is ready .

Mahi0911:
If one of the database pods which is initialising database , is killed/stopped before it's ready , the other pods try to create the database and result in Pod Crash  #1969 
Solved this by creating a checkpoint file and checking if it is present before creating or starting database.


Mahi0911:
#1971

Mahi0911:
Addressed the issue #1972 by changing permissions of "`$ORACLE_BASE/scripts/extensions/setup/*.sh`" file

yunus-qureshi (approved):
Looks fine, thanks

gvenzl (approved):

gvenzl:
Awesome, thanks @Mahi0911!

Issue #1974, opened 2021-06-14
Helm charts help text displays wrong $NODE_IP 


Mahi0911:
Help text displays <nodeport-ip> as $NODE_IP even when loadBalService is set to true in values.yaml .

PR #1975, opened 2021-06-14
Modify help text displayed during helm install

Mahi0911:
Help text displays <nodeport-ip> as $NODE_IP even when loadBalService is set to true in values.yaml . #1974 
Modified the same to display <loadBalancer-ip> as $NODE_IP when loadBalService is set to true in values.yaml


yunus-qureshi (approved):
Looks fine to me

gvenzl (approved):

PR #2007, opened 2021-07-06
Fixing issues in 19c

### LINKED ISSUE #2006: Unwanted data in logs and tnsnames.ora content overwriting issue
When database image is extended with k8s extension, some unwanted numerical data appears in the log. This is actually output of pgrep command used in the scripts.
Also if we set tnsnames.ora before dbca command execution, tnsnames.ora get overwritten by dbca. So, we lose tns entry for PDB. This further creates problem in cloning database in kubernetes environment.

abhisbyk:
Close #2006 

Fixed following issues:

- Removal of unwanted data appearing in logs after building k8s extension.
- Fixing tnsnames.ora overwriting issue.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/19.3.0/createDB.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/19.3.0/createDB.sh
@@ -15,7 +15,7 @@
 
 set -e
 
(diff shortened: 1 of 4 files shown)


yunus-qureshi (approved):
Seems fine

gvenzl (approved):

PR #2009, opened 2021-07-07
Adding support for additional init parameters in 19c

### LINKED ISSUE #2008: Support for cpu_count and processes database init parameters in 19c container image
Till now only two init parameters are supported: `INIT_SGA_SIZE` and `INIT_PGA_SIZE`.
Support is required for setting up cpu_count and processes while creating the database.

abhisbyk:
Close #2008 

Added env var `INIT_PARAMS`. It takes a string of the following format:
**"\<key1\>=\<value1\>,\<key2\>=\<value2\>..."**
where `key` is the init parameter and `value` is the value for that parameter.

When this env var is passed to `docker run` command, it sets the specified init parameters of the database.

Signed-off-by: abhisbyk <abhishek.by.kumar@oracle.com>

### DIFF
--- a/OracleDatabase/SingleInstance/README.md
+++ b/OracleDatabase/SingleInstance/README.md
@@ -74,6 +74,7 @@ To run your Oracle Database image use the `docker run` command as follows:
     -e ORACLE_EDITION=<your database edition> \
     -e ORACLE_CHARACTERSET=<your character set> \
     -e ENABLE_ARCHIVELOG=true \
+    -e INIT_PARAMS=<additional database init parameters of format "key1=value1,key2=value2..."> \
     -v [<host mount point>:]/opt/oracle/oradata \
     oracle/database:19.3.0-ee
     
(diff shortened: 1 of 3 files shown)


yunus-qureshi (approved):
Thanks. This is a useful option

gvenzl:
A couple of things:

1) If we were to introduce this, should we not retire `INIT_SGA_SIZE` and `INIT_PGA_SIZE`?
2) What's wrong with the current approach for users to modify parameters and anything else they like via the[`/opt/oracle/scripts/setup`](https://github.com/oracle/docker-images/tree/main/OracleDatabase/SingleInstance#running-scripts-after-setup-and-on-startup) scripts?
3) What's the actual business case for this? In the last 6 years, this has not been requested yet. What has changed? 

Some additional thoughts: we have 453 parameters in 19c, some of them cannot be changed, others when set to invalid values will prevent the database creation altogether. When the database creation fails, users will have no easy way to find out what failed and/or why. They will have to get into the container, find the correct log and trace files and inspect them without having traditional programs available such as `less` or `vim`. None of that is very container-friendly nor what users would expect from an image. By having explicit environment variables, the entry point script itself can validate and guarantee that the value is correct, or choose to ignore it altogether, etc. These have all been reasons why we never opened up parameter support lightly for these images, only what's really necessary.

yunus-qureshi:
@gvenzl

Me and @abhisbyk had discussion on this. The actual use case is for sharding where cpu_count and processes init params need to be bumped up on certain nodes. 
Also we want to be able to support this for K8s deployments where running setup/start scripts is not easily doable.

But we tend to agree with your concerns that opening this up to modify any init param can cause a number of unexpected issues. 
We'll close this PR and open a new one adding  just INIT_CPU_COUNT and INIT_PROCESSES

PR #2020, opened 2021-07-15
RFC: switch to using the full oraclelinux:8 base image for the Oracle Linux Developer images

Djelibeybi:
This makes the developer images far more developer-friendly by including such useful tools as a module-aware `dnf` and more. Yes, the images jump in size, but then developer images are inherently larger because of their purpose.

Of particular importance is for folks to test the builds at https://github.com/Djelibeybi?tab=packages to make sure current functionality is unaffected, including any downstream CI/CD tooling that assumes the existence of `microdnf`.


crush-157 (approved):
The `2.7-node` image seems fine.  I've hit an issue with compiling native gem extensions in the `2.7` image, but it's consistent with the current (official) image, so I'll look at this once we've fattened the images.

Djelibeybi:
> I've hit an issue with compiling native gem extensions in the `2.7` image, but it's consistent with the current (official) image, so I'll look at this once we've fattened the images.

That would be great.

AmedeeBulle:
Can you clarify how you would like to transition to `oraclelinux:8` beyond this RFC?

Is the plan to re-base all dev images on the `compat` image and as second step have the code owners move from `microdnf` to `dnf` and drop the dependency on `compat`?

For the _migrated_ images (like your `golan/latest`)  shouldn't we drop the module file?

Djelibeybi:
No, we have to maintain `microdnf` and the module files until at least the next Oracle Linux update is released, to give downstream consumers of the images time to update their tooling. My plan is actually to provide two updates cycles for this. 

When the next update is released, my plan is to replace `microdnf` with a script that outputs a notice saying the image has changed and to switch to `dnf` before the next update is released, at which point `microdnf` will be removed from this image. Then the script will pause for 30 seconds. That pause should be sufficient for folks to notice their CI jobs are taking longer and to check the logs. 

At the next update, we can finalise the switch by removing the module files and switching directly to the `oraclelinux:8` base image.

OracleLinuxDevelopers/oraclelinux8/nodejs/14-oracledb/Dockerfile
```diff
@@ -1,15 +1,17 @@
...
-RUN microdnf install oracle-instantclient-release-el8 oraclelinux-developer-release-el8 && \
-    microdnf install libaio \
-                     oracle-instantclient-basic \
-                     node-oracledb-node14 \
-                     && \
-    microdnf clean all
+RUN dnf -y install oracle-instantclient-release-el8 \
+                   oraclelinux-developer-release-el8 \
+                   && \
+    dnf -y install libaio \
```
cjbj, line 10: The IC Basic package has a dependency on libaio, so technically this explicit dependency isn't need.

OracleLinuxDevelopers/oraclelinux8/golang/latest/Dockerfile
```diff
@@ -1,10 +1,10 @@
 # Copyright (c) 2020 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-FROM oraclelinux:8-slim
+FROM ghcr.io/djelibeybi/oraclelinux8-compat:8-slim
```
cjbj, line 3: To end users this 'strange' repo would looks sus.  Can you add some comments to each Dockerfile to explain what it is, and what the plan is?

cjbj (changes requested):

OracleLinuxDevelopers/oraclelinux8/golang/latest/Dockerfile
Djelibeybi: Whoops, that was for testing purposes. I need to merge the compat image into this repo and change the URL. It will be `ghcr.io/oracle/oraclelinux8-compat:8-slim` eventually.

Djelibeybi:
Once #2044 is merged and the `oraclelinux8-compat:8-slim` image is built, I will update this PR to use that image from Oracle's GHCR.

Djelibeybi:
@AmedeeBulle I would deeply appreciate it if you could fix the workflow breakage I caused. :)

OracleLinuxDevelopers/oraclelinux8/nodejs/14-oracledb/Dockerfile
Djelibeybi: I've left the original Dockerfiles intact, i.e. I just want to change the base image. We can fix this afterwords.

Djelibeybi:
@AmedeeBulle @mark-au @totalamateurhour can I please get another review/approval from one of you? I'd like to merge this sooner rather than later.

OracleLinuxDevelopers/oraclelinux8/nodejs/14-oracledb/Dockerfile
```diff
@@ -1,15 +1,17 @@
 # Copyright (c) 2021 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-FROM oraclelinux:8-slim
+FROM ghcr.io/oracle`/oraclelinux8-compat:8-slim
```
cjbj, line 3: Stray backquote

cjbj (dismissed):
I checked the two with DB drivers; they build and run fine.

- Can/should you augment the slim note to say whether users should migrate from microdnf?
- Technically you can remove the explicit install of oracle-instantclient-basic since that gets pulled in automatically.  I'm in two minds about 'hiding' this, since the result is the same.
- There is at least one changed file with only a 2020 copyright.

Djelibeybi:
Thanks @cjbj, I've pushed changes to the main `README.md` with a deprecation notice for `microdnf` and I've updated all the applicable copyright notices. I'm going to leave the explicit installation of `oracle-instantclient-basic` because there's no change in the result and it makes it more obvious that it's part of the image when only reviewing the source.

Djelibeybi:
@AmedeeBulle @totalamateurhour @mark-au it would be great to get another review, but I will be merging this change in regardless by the end of the week, because we need to stop using `8-slim`.

AmedeeBulle (approved):

Issue #1544, opened 2020-03-09
Oracle database 18 XE container fails to start on Centos


taspolle8:
Hello,

First of all, unfortunately I am unable to post full logs due to the host being in a closed network as well as security issues, but I will try my best to find all the nesessary info.

Host: fresh CentOS 7 install
Image: Oracle database 18.4.0 XE
Docker version: 19.03.7 build 7141c199a2

Image was originally built on a windows machine and transferred to closed network Windows 10 host, where running it worked without problems. 
Transfering it on the linux host is where i ran into problems (I should be able to run the linux image on both, i assume).

```
$docker run --name xx -d -p 1521:1521 -p 5500:5500 qwerty:asdfg
-- Container starts up

Inside container though:
$lsnrctl status
TNS-12541 TNS:no listener
TNS: protocol adapter error
No listener
Linux Error: 111: Connection Refused.

$docker logs
Oracle base remains unchanged with value /opt/oracle
su: cannot open session: permission denied
#################################
DATABASE SETUP WAS NOT SUCCESSFUL!
```

From what I tried to google and find, SELinux might be a problem, but after setting it disabled, there is not change.
```
$sestatus
SELinux status: Disabled
```

Also tried adding the tmpfs line on /etc/fstab on centos machine

```
$vi /etc/fstab
UUID=ce2fce41-asdf2-...- /boot
/dev/mapper/centos-swap swap   swap  defaults  0 0
tmpfs    /dev/shm     tmpfs   defaults,size=1024m 0 0
```

(Edited for formatting)
Any help would be much appreciated, thanks!

Djelibeybi:
That last line in your `/etc/fstab` is wrong (there shouldn't be a `$` at the beginning). Can you paste the actual content of the `/etc/fstab` from your CentOS machine? 

Also, I would strongly encourage you to install Oracle Linux 7 and use Oracle Container Runtime for Docker to both build and run the image, just so we can eliminate any weirdness that may have come from building the image on Windows. I don't _think_ that's the problem, but I can't rule it out either.

We also need the output from `docker info` and `docker logs` (sanitised is fine) from your CentOS box because the `su` error is a known issue with XE and there is probably/hopefully something else in the logs that indicates the actual issue, like a filesystem issue somewhere.

taspolle8:
Thank you very much Djelibeybi for a quick reply!

ah yes the $ in fstab is a typo i typed here. 

```
$docker info
Containers: 1
Running: 1
Paused: 0
[...]

PR #2024, opened 2021-07-19
Fix for CentOS, Amazon Linux

artemiuss:
published here: https://github.com/oracle/docker-images/issues/1544#issuecomment-740570620


AmedeeBulle (commented):
It would be interesting to know which limit is actually exceeded.

The proposed fix disables/bypass all limits which might defeat the purpose of the check.

If we are OK with that, I am OK to merge...
(the next line is then probably redundant...)

Djelibeybi:
Closing due to lack of reponse.

PR #2042, opened 2021-08-04
Handled Timing issue with acquiring/releasing locks

Mahi0911:
Enhanced lock.py by removing unnecessary checks causing the timing issue .


yunus-qureshi (approved):
Looks fine, thanks

gvenzl (approved):

PR #2044, opened 2021-08-04
feat(dev-images): create a base image for compatibility purposes

Djelibeybi:
This image adds microdnf to the oraclelinux:8 image so that any
downstream automation continues to work after the replacement of
the current oraclelinux:8-slim base image.

This the first step and is required to update #2020 to reference 
the right base image.

### DIFF
--- a/.github/workflows/build-and-push-dev-images.yml
+++ b/.github/workflows/build-and-push-dev-images.yml
@@ -27,13 +27,13 @@ on:
         required: false
       lang:
         description: List of languages to build
-        default: 'golang, nodejs, php, python, ruby'
+        default: 'compat, golang, nodejs, php, python, ruby'
         required: false
 
 # Default values for the builds triggered by the push event
 env:
   ol: 'oraclelinux7, oraclelinux8'
-  lang: 'golang, nodejs, php, python, ruby'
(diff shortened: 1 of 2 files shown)


mark-au (approved):
If the upstream image is changing I guess we need this compatibility image 👍

.github/workflows/build-and-push-dev-images.yml
```diff
@@ -27,13 +27,13 @@ on:
         required: false
       lang:
         description: List of languages to build
-        default: 'golang, nodejs, php, python, ruby'
+        default: 'compat, golang, nodejs, php, python, ruby'
```
AmedeeBulle, line 30: This introduce dependencies when images start using compat.
A change in compat should trigger a rebuild for for the compat based, similarly a change in _upstream_ will require a compat rebuild beforehand.

I would suggest to get this PR through now and tackle the potential rebuild issues in a separate PR...

AmedeeBulle (approved):

.github/workflows/build-and-push-dev-images.yml
Djelibeybi: That makes sense.

PR #2045, opened 2021-08-04
fix(doc): Update the contribution guide to be more specific

Djelibeybi:
We do not recommend using oraclelinux:8-slim for any image that
needs to install additional packages.


mark-au:
I didn't know we explicitly discouraged installing additional RPMs with `microdnf` but will take your lead on that. Agree that the change does what you intend.

mark-au (approved):
I didn't know we explicitly discouraged installing additional RPMs with microdnf but will take your lead on that. Agree that the change does what you intend.

Djelibeybi:
`microdnf` is not module-aware, so it can lead to unintended consequences when used. We're trying to avoid that ambiguity.

PR #2055, opened 2021-08-14
Cleaned the yum cache after the yum update for ServerJRE8

aureliogrb:
Also removed unneeded clean up from the builder image and changed from yum/microdnf clean all to rm -rf /var/cache/yum 

Signed-off-by: Aurelio Garcia-Ribeyro <aurelio.garciaribeyro@oracle.com>

### DIFF
--- a/OracleJava/11/Dockerfile.8-slim
+++ b/OracleJava/11/Dockerfile.8-slim
@@ -50,8 +50,7 @@ COPY $JAVA_PKG /tmp/jdk.tgz
 RUN set -eux; \
 	echo "$JAVA_SHA256 */tmp/jdk.tgz" | sha256sum -c -; \
 	mkdir -p "$JAVA_HOME"; \
-	tar --extract --file /tmp/jdk.tgz --directory "$JAVA_HOME" --strip-components 1; \
-	rm /tmp/jdk.tgz; 
+	tar --extract --file /tmp/jdk.tgz --directory "$JAVA_HOME" --strip-components 1
 	
 ## Get a fresh version of SLIM for the final image	
 FROM oraclelinux:8-slim
@@ -73,7 +72,7 @@ RUN set -eux; \
 # JDK assumes freetype is available		
 		freetype fontconfig \
 	; \
-	microdnf clean all; \
+	rm -rf /var/cache/yum; \
 	ln -sfT "$JAVA_HOME" /usr/java/default; \
 	ln -sfT "$JAVA_HOME" /usr/java/latest; \
 	for bin in "$JAVA_HOME/bin/"*; do \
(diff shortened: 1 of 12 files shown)


Djelibeybi:
On Oracle Linux 8, you need to run `rm -rf /var/cache/dnf` instead.

OracleJava/11/Dockerfile.8-slim
```diff
@@ -73,7 +72,7 @@ RUN set -eux; \
 # JDK assumes freetype is available		
 		freetype fontconfig \
 	; \
-	microdnf clean all; \
+	rm -rf /var/cache/yum; \
```
Djelibeybi, line 75: This should be `/var/cache/dnf`

OracleJava/8/Dockerfile.8-slim
```diff
@@ -70,6 +69,7 @@ COPY --from=builder $JAVA_HOME $JAVA_HOME
 
 ##
 RUN	microdnf -y update; \
+	rm -rf /var/cache/yum; \
```
Djelibeybi, line 72: This should be `/var/cache/dnf`

Djelibeybi (dismissed):
You should switch to using the `oraclelinux:8` image and change from `microdnf` to `dnf` as well. We only recommend using the `8-slim` image with microservices that are statically compiled and don't require any additional packages.

OracleJava/8/Dockerfile.8-slim
aureliogrb: Fixed.  Thanks

OracleJava/11/Dockerfile.8-slim
aureliogrb: Fixed.  Thanks

aureliogrb: I will use OL8 for JDK 17+ but will keep the current images on OL8 Slim to avoid surprising existing users.

Djelibeybi (changes requested):
You still need to change `FROM oraclelinux:8-slim` to `FROM oraclelinux:8` in all the OL8-based images.

aureliogrb:
I will change to oraclelinux:8 for JDK 17 and later but don't think we should do so for the existing images that are already being offered in 8-slim.

Djelibeybi:
You can change the existing images to `FROM ghcr.io/oracle/oraclelinux8-compat:8-slim` so that they get the compatibility base image. Essentially, this is `oraclelinux:8` with both `dnf` and `microdnf` installed so any downstream CI/CD that relies on `microdnf` will continue to work.

aureliogrb:
Updated the oraclelinux:8-slim references to ghcr.io/oracle/oraclelinux8-compat:8-slim.  Since that image has both dnf and microdnf changed scripts to use dnf.  Also removed some unnecessary echos at the end of the build file and unneded xshare dump for versions > 11.

Djelibeybi (approved):

PR #2085, opened 2021-08-30
feat(developper):  add golang 1.16 container

AmedeeBulle:
Add a a golang 1.16 specific container. For now it will generate the
same container as the one in the `golang` directory, but it "locks" the
version on 1.16.
It will also be built by the `build-and-push-dev-images` action.
Also update containers from previous golang version to disable any
pre-set golang channel. This will avoid issue if/when 1.16 is no more
the default version.


totalamateurhour (approved):
LGTM

Djelibeybi (approved):

PR #2089, opened 2021-09-01
fix: remove the module files as they are created by dnf

Djelibeybi:
@AmedeeBulle (gently) pointed out that we no longer need the `.module` files as `dnf module enable` creates them.


AmedeeBulle (approved):
Massive cleanup!

LGTM

Djelibeybi:
I suddenly feel so light and airy. .:)

Issue #2048, opened 2021-08-07
No oracle user, oracle group in oracle/fmw-infrastructure:12.2.1.4


jmalbarran:
The install procedure asumes that there is an oracle user, oracle group in base image oracle/fmw-infrastructure:12.2.1.4. But there is not

It fails in STEP 9 of docker build, inside 

`sh buildDockerImage.sh -v 12.2.1.4`

Output:
```
$ sh buildDockerImage.sh -v 12.2.1.4
sh: buildDockerImage.sh: No such file or directory
[oracle@wcs-docker-server 12.2.1.4]$ cd ..
[oracle@wcs-docker-server dockerfiles]$ sh buildDockerImage.sh -v 12.2.1.4
Checking if required packages are present and valid...
V983400-01.zip: OK
=====================
Building image 'oracle/wcsites:12.2.1.4' ...
Building image command: docker build --force-rm=true --no-cache=true  -t oracle/wcsites:12.2.1.4 -f Dockerfile
STEP 1: FROM oracle/fmw-infrastructure:12.2.1.4 AS builder
STEP 2: USER root
--> 02f7d204372
STEP 3: ENV SITES_CONTAINER_SCRIPTS=/u01/oracle/sites-container-scripts 	SITES_INSTALLER_PKG=wcs-wls-docker-install 	SITES_PKG=V983400-01.zip 	SITES_JAR=fmw_12.2.1.4.0_wcsites.jar
--> 5d24d080e9d
STEP 4: COPY sites-container-scripts/* $SITES_CONTAINER_SCRIPTS/
--> 8108c7a4453
STEP 5: COPY $SITES_INSTALLER_PKG /u01/$SITES_INSTALLER_PKG
--> 9c38ad8ed4d
STEP 6: COPY $SITES_PKG install.file oraInst.loc /u01/
--> 7ef79aa45fe
STEP 7: COPY patches/* /u01/patches/
--> e4616cfdc04
STEP 8: RUN mkdir -p /u01/oracle/logs && 	chmod a+xr /u01 && 	chown oracle:oracle -R /u01 && 	chmod a+xr -R /u01/oracle/*.* && 	chmod a+xr $SITES_CONTAINER_SCRIPTS/*.*
chown: invalid group: 'oracle:oracle'
STEP 9: FROM oracle/fmw-infrastructure:12.2.1.4
[...]

PR #2094, opened 2021-09-07
Handle creation of oracle user and group on fmw-infrastructure in WCS docker file

tfield:
This fixes https://github.com/oracle/docker-images/issues/2048 and ensures that the installer can be built for WCS.


Djelibeybi:
Closed due to lack of response.

PR #2162, opened 2021-10-15
Updated PATH env variable to absolute value

MarthaMahipal:
Updated PATH env variable to absolute value due to security reasons


Mahi0911:
This PR addresses issue : https://github.com/oracle/docker-images/issues/2163

Mahi0911:
This was an argument from Security team , that If Someone alters the PATH in base image, it would be a security concern.

Djelibeybi:

PR #2170, opened 2021-10-21
Correcting the position of umask reversion for clone/standby db

### LINKED ISSUE #2169: PDB connectivity issue with the Clone Database
When oracle wallet is used to create the database, Database PDB connection fails with SQLPlus. This is caused by not reverting file creation mask for this specific use case.

abhisbyk:
Close #2169 

Correction reversion of file creation mask (using `umask 022`) for clone database usecase, when Oracle wallet is used.

Signed-off-by: abhisbyk <abhishek.by.kumar@oracle.com>


Mahi0911:
Seems fine to me

PR #2183, opened 2021-11-11
Fix Node.js 14 Dockerfile

Shogobg:
There was a missing closing bracket for the command in the Node.js 14 Dockerfile.


Djelibeybi (approved):

Djelibeybi:
Thanks, good catch.

PR #2196, opened 2021-12-01
feat: add gcc-toolset-11 toolchain and perftool images.

### LINKED ISSUE #2185: GCC-Toolset images
With the release of Oracle Linux 8.5 we should see provision for the gcc-toolset 11. Previously, I think on 8.4 we had gcc-devtoolset 9. It would be good if there were images available for these tools, as Red Hat provide the following:

    rhel8/gcc-toolset-11-toolchain, which includes the GCC compiler, the GDB debugger, and the make automation tool.
    rhel8/gcc-toolset-11-perftools, which includes the performance monitoring tools, such as SystemTap and Valgrind.

Djelibeybi:
Test builds of the images are available via my fork at https://github.com/Djelibeybi/docker-images/pkgs/container/oraclelinux8-gcc-toolset. 

Resolves #2185.

### DIFF
--- a/.github/workflows/build-and-push-dev-images.yml
+++ b/.github/workflows/build-and-push-dev-images.yml
@@ -29,14 +29,14 @@ on:
         required: false
       lang:
         description: List of languages to build
-        default: 'compat, golang, nginx, nodejs, php, python, ruby'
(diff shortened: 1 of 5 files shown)


AmedeeBulle (approved):
LGTM

PR #2238, opened 2022-01-25
Auto increase DBCA total memory based on the container memory

felipecrs:
This introduces a smarter memory allocation feature for DBCA, allowing it to expand to the allocated memory for the container (based on cgroups).

Previously, no matter how much memory you allocate to the Oracle container, the database would never use more than 2GB. You could tweak it at some sort by using https://github.com/oracle/docker-images/pull/2239, but that isn't convenient, neither portable/scalable. Not to mention that it requires manual (re-)calculation of values.

So, if the user allocates 3GB for the container (`docker run --memory 3g`), it will be used. Previously, to make use of more than 2GB, the user would have
to:

1. Know that the default configuration is limited to 2GB. This is already something that most users would not know.
2. Allocate more than 2GB by tweaking INIT_SGA_SIZE and INIT_PGA_SIZE, which in turn would require some manual calculations as well (which is always error prone).

Some more caveats are:

3. The user would need to remember to re-do these calculations every time the memory allocated for the container is changed.

---

Also solves the following bug:

If you use a big environment (more than 8 CPUs) to build the prebuiltdb extension,
the image generated by it won't be portable to other systems.

This makes the `createDB.sh` script aware of an environment variable that would
skip "optimization" for big environments, and thus making the result image portable.

This is very reasonable to be the default in the prebuiltdb extension,
otherwise, users (like me) would end up with difficult issues to be debugged.
Unless the person who is executing the prebuiltdb has extensive knowledge about
how the Oracle DB initialization scripts works.

For very advanced use cases, one can simply pass `--build-arg INIT_DONT_OPTIMIZE_FOR_BIG_ENVIRONMENTS=false` which I believe does not even worths
to be documented.

Closes #1989

PR #2239, opened 2022-01-26
Add INIT_SGA_PERCENTAGE for smarter memory allocation

felipecrs:
This is another take for the issue reported in https://github.com/oracle/docker-images/pull/2238.

I believe this is a much better and cleaner approach than the one used in https://github.com/oracle/docker-images/pull/1742, and that's why it replaces the existing one.

Furthermore, now the default configuration is smarter. If the user allocates 3GB for the container, it will be used. Previously, to make use of more than 2GB, the user would have
to:

1. Know that the default configuration is limited to 2GB. This is already something that most users would not know.
2. Allocate more than 2GB by tweaking INIT_SGA_SIZE and INIT_PGA_SIZE, which in turn would require some manual calculations as well (which is always error prone).

Some more caveats are:

3. The user would need to remember to re-do these calculations everytime the memory allocated for the container is changed.

### DIFF
--- a/OracleDatabase/SingleInstance/README.md
+++ b/OracleDatabase/SingleInstance/README.md
@@ -70,8 +70,7 @@ To run your Oracle Database image use the `docker run` command as follows:
     -e ORACLE_SID=<your SID> \
     -e ORACLE_PDB=<your PDB name> \
     -e ORACLE_PWD=<your database passwords> \
-    -e INIT_SGA_SIZE=<your database SGA memory in MB> \
-    -e INIT_PGA_SIZE=<your database PGA memory in MB> \
+    -e INIT_SGA_PERCENTAGE=<memory percentage of your database SGA> \
     -e ORACLE_EDITION=<your database edition> \
     -e ORACLE_CHARACTERSET=<your character set> \
     -e ENABLE_ARCHIVELOG=true \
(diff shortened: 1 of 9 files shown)


felipecrs:
Now, just out of curiosity, why do we bother tweaking SGA/PGA values instead of just tweaking the total memory?

abhisbyk:
@felipecrs `INIT_SGA` and `INIT_PGA` are honoured when the database is created. If a pre-created database is started again then `INIT_SGA` and `INIT_PGA` params are not used. The situation with `INIT_SGA_PERCENTAGE` is also the same. 

So, in the case of **prebuiltdb** extension, the database is created at the build time and then added as an additional layer on top of the base image. So, neither `INIT_SGA/INIT_PGA` nor `INIT_SGA_PERCENTAGE` would work if a database is started using the **prebuiltdb** extended image. Hence, the **prebuiltdb** image will not be portable after building on the large envrionments.

Also, we have other products(e.g. OraOperator) that uses the INIT_SGA/INIT_PGA params of SIDB image, so changing that to INIT_SGA_PERCENTAGE will harm this dependence.

PR #2238 will definitely be helpful in this context.

felipecrs:
Got it, thanks!

PR #2240, opened 2022-01-26
Automatically use the allocated memory

felipecrs:
This is another take on https://github.com/oracle/docker-images/pull/2239, which does not remove the existing INIT_SGA_SIZE and INIT_PGA_SIZE parameters, but still solves the issue reported at https://github.com/oracle/docker-images/pull/2238 (because now the total memory configuration is re-assessed on every container start).

And also, it brings all the benefits listed on https://github.com/oracle/docker-images/pull/2239 about the memory allocation.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/19.3.0/createDB.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/19.3.0/createDB.sh
@@ -165,18 +165,11 @@ sed -i -e "s|###ORACLE_CHARACTERSET###|$ORACLE_CHARACTERSET|g" "$ORACLE_BASE"/db
 
 # If both INIT_SGA_SIZE & INIT_PGA_SIZE aren't provided by user
 if [[ "${INIT_SGA_SIZE}" == "" && "${INIT_PGA_SIZE}" == "" ]]; then
-    # If there is greater than 8 CPUs default back to dbca memory calculations
-    # dbca will automatically pick 40% of available memory for Oracle DB
-    # The minimum of 2G is for small environments to guarantee that Oracle has enough memory to function
-    # However, bigger environment can and should use more of the available memory
-    # This is due to Github Issue #307
-    if [ "$(nproc)" -gt 8 ]; then
-        sed -i -e "s|totalMemory=2048||g" "$ORACLE_BASE"/dbca.rsp
-    fi;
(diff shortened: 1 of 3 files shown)


abhisbyk:
@felipecrs The total memory configuration **will not be re-assessed** on every container restart, as the **prebuiltdb** extended image has **a pre-created database**. If this image is used to start the database, neither `INIT_SGA/INIT_PGA` params will be honoured, nor `ALLOCATED_MEMORY` will calculate the newer memory allocated to the container.

These params are used when the DB image is run for the first time(i.e. when the database is created), which is happening while building the **prebuiltdb** extension (i.e. docker build-time). If the **prebuiltdb** extension is built on the "**big environments**" the `ALLOCATED_MEMORY` env var will be very large while building the image, and this will stay with the **prebuiltdb** extended image, because of the above mentioned reason. This would be similar to the non-portable image condition that you've mentioned.

PR #2238 will definitely be useful in this context.

felipecrs:
Alright. But what if we move this code from `createDb.sh` to `runOracle.sh`, would it help to be able to assess the total available memory on every container start?

abhisbyk:
Init params are passed through dbca.rsp (DBCA response file) while creating the database. No matter where we put the code, these params will be passed to the database only at the creation of the database.

felipecrs:
Don't you think that this PR when combined with https://github.com/oracle/docker-images/pull/2238 (to fix prebuiltdb) would be a lot benefitial?

Instead of relying on the number of CPUs, we rely on the available memory. I have VMs with 8vCPUs and 64GB of RAM, but Oracle won't use more than 2GB unless the VM has 8vCPUs.

Perhaps we can tweak this PR a bit, like instead of making it the new default behavior, one can opt-in to do that.

felipecrs:
PS: this (https://github.com/oracle/docker-images/pull/2238#issuecomment-1027912412) could also have the same effect as this PR, but it depends on how DBCA calculates the available memory. For containers, it's important to take into account the memory information from `cgroup` and `cgroupv2`. I know that recent versions of JVM make use of it.

So, if DBCA respects cgroup/cgroupv2, https://github.com/oracle/docker-images/pull/2238#issuecomment-1027912412 may supersede this PR. If not, I still think this PR would be a great addition.

Issue #1594, opened 2020-04-26
How to reach noSQL outside the container


olelbis:
I'm able to use CLI, Ping and also sql for noSQL inside a docker.
But whe i try for example:
docker run -h store -d -p 5000:5000 --name=kvlite oracle/nosql

on my windows machine:
netstat -an|find "5000"
  TCP    0.0.0.0:5000           0.0.0.0:0              LISTENING
  TCP    [::]:5000              [::]:0                 LISTENING

from cmd or powershell i try:
D:\Java\jdk8\bin\java.exe -jar .\kvstore.jar ping -host localhost -port 5000
Can't find store topology: Could not contact any RepNode at: [localhost:5000]

Same results if i try with my hostname or ip address.

Cheers

adeywong:
Bumping this - anyone ?

Djelibeybi:
Have you opened the firewall on your Docker host to allow port 5000?

adeywong:
I have disabled my firewall just in case, but that does not seem to be the issue.
Has anyone been able to successfully connect from their host to the OracleNoSql inside a container? There was recommendation was to run the container with host networking mode, but this doesn't work if the host is a MacOS. I have also modified the dockerfile to start kvlite with -host 0.0.0.0 and it didnt work either.

adeywong:
Also want to note that container to container connection as per the example on the docker page works fine
```
 docker run --rm -ti --link kvlite:store oracle/nosql \
  java -jar lib/kvstore.jar ping -host store -port 5000
```

olelbis:
> Also want to note that container to container connection as per the example on the docker page works fine
> 
> ```
>  docker run --rm -ti --link kvlite:store oracle/nosql \
>   java -jar lib/kvstore.jar ping -host store -port 5000
> ```

Exactly, i’also able to use cotainer to container connection succesfully.

But if i try from my linux or win machine, i got topology error.

If i install nosql locally instead using docker everithing work fine.

NoSQLGuy:
The IP address for the docker container running the store is different. To get the container ip address, do the following:
Get the docker container ip address. Issue this command immediately after creating the container : docker inspect nosql-container | grep -i ipadd
 
You can then ping the store using the container IP address:
 
docker run --name nosql-container -d kvlite:latest
docker exec -it nosql-container bash
 
On the docker bash prompt, issue the following command:
java -jar lib/kvstore.jar ping -port 5000 -host <container_ip_address>

[...]

PR #2246, opened 2022-02-04
Update the NoSQL Image

dario-vega:
This PR updates with some improvements to the Oracle NoSQL CE image.
    Provide a solution to https://github.com/oracle/docker-images/issues/1594 - ADVANCED CONFIGURATION
    Avoid installing new packages - not using anymore `microdnf install unzip hostname`
    When restarting the container, KVLITE refuses to start because of the hostname change. Use -hostname when executing the command run.

Finally, I updated build-and-push-nosql-image.yml
    moving from a release-based tag to a date-based tag
    changing the paths
	
Signed-off-by: Dario Vega dario.vega@oracle.com

### DIFF
--- a/NoSQL/README.md
+++ b/NoSQL/README.md
@@ -162,7 +162,7 @@ In this case, You need to publish all internal ports and the KV_PROXY_PORT.
 - 8080 KV_PROXY_PORT
 
 This hostname must be resolvable from the host outside the container. 
-It could be an alias to the host running the docker commands (e.g nosql-container-host).
+It could be an alias to the host running the docker commands.
 
 ```shell
 $ cat /etc/hosts
--- a/NoSQL/README.md
+++ b/NoSQL/README.md
@@ -38,7 +38,7 @@ The steps outlined below are using Oracle NoSQL Database community edition, if
 you are using Oracle NoSQL Database Enterprise Edition, please use the
 appropriate image name.
 
-Start up KVLite in a Docker container. You must give it a name and provide a hostname. Startup of
+Start up KVLite in a container. You must give it a name and provide a hostname. Startup of
 KVLite is the default `CMD` of the image:
 
 ```shell
@@ -111,7 +111,7 @@ $ docker run --rm -ti --link kvlite:store oracle/nosql:ce \
 
 ## Using Oracle NoSQL Command-Line from an external host
 
-Start up KVLite in a Docker container. You must give it a name 
+Start up KVLite in a container. You must give it a name 
 and provide a hostname (resolvable from the external host - an alias to the container host). 
 You need to publish all internal ports.
 
@@ -128,15 +128,16 @@ instance:
 $ cat /etc/hosts
 10.0.0.143 nosql-container-host
 10.0.0.143 kvlite-nosql-container-host
-
-
+```
+```shell
 $ ping kvlite-nosql-container-host
 
 PING kvlite-nosql-container-host (10.0.0.143) 56(84) bytes of data.
 64 bytes from nosql-container-host (10.0.0.143): icmp_seq=1 ttl=64 time=0.259 ms
 64 bytes from nosql-container-host (10.0.0.143): icmp_seq=2 ttl=64 time=0.241 ms
 64 bytes from nosql-container-host (10.0.0.143): icmp_seq=3 ttl=64 time=0.192 ms
-
+```
+```shell
 $ java -jar $KVHOME/lib/kvstore.jar ping -host kvlite-nosql-container-host -port 5000
 ```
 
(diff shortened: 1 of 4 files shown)


dario-vega:
Avi, can you test the solution for https://github.com/oracle/docker-images/issues/1594 as discussed, please? (Mac)
Read Using Oracle NoSQL Command-Line from an external host
It is considered an advanced configuration for us.

NoSQL/README.md
```diff
@@ -103,6 +109,62 @@ $ docker run --rm -ti --link kvlite:store oracle/nosql:20.3 \
...
+
+Start up KVLite in a Docker container. You must give it a name 
+and provide a hostname (resolvable from the external host - an alias to the container host). 
+You need to publish all internal ports.
+
+Startup of KVLite is the default `CMD` of the image:
+
+```shell
+docker run -d --name=kvlite --hostname=kvlite-nosql-container-host --env KV_PROXY_PORT=8080 -p 8080:8080 \
+-p 5000:5000 -p 5010-5020:5010-5020 -p 5021-5049:5021-5049 -p 5999:5999 oracle/nosql:ce
```
Djelibeybi, line 122: You can combine the second and third port mappings to `-p 5010-5049:5010-5049` as they are contiguous.

NoSQL/ce/start-kvlite.sh
```diff
@@ -1,7 +1,12 @@
...
-KVHOST=`hostname`
-java -jar lib/kvstore.jar kvlite  -secure-config disable -root /kvroot &
-java -jar lib/httpproxy.jar -helperHosts $KVHOST:5000 -storeName kvstore -httpPort $KV_PROXY_PORT -verbose true
++set -ex
+
+KVHOST=$HOSTNAME
+_KV_HARANGE=`echo $KV_HARANGE | sed 's/\-/\,/g' `
+_KV_SERVICERANGE=`echo $KV_SERVICERANGE | sed 's/\-/\,/g' `
+java -jar lib/kvstore.jar kvlite -secure-config disable -root /kvroot -host $KVHOST -port $KV_PORT -admin-web-port $KV_ADMIN_PORT -harange $_KV_HARANGE -servicerange $_KV_SERVICERANGE &
+java -jar lib/httpproxy.jar -helperHosts $KVHOST:$KV_PORT -storeName kvstore -httpPort $KV_PROXY_PORT -verbose true
```
Djelibeybi, line 12: Also, I had to insert a loop between the two commands so that the proxy didn't exit and shutdown the container prematurely. I went with this:

```bash
while [ ! -d /kvroot/kvstore/log ]; do
    echo "Waiting for kvstore to start..."
    sleep 1
done
```

Which seems to be enough. If there's a better way to get the proxy process not to give up so quickly, use that. :)

Djelibeybi (changes requested):
Oracle NoSQL seems extremely picky about the hostname used to connect from a client to the container running `kvlite`. Also, just opening port 5000 isn't enough, the service ports need to be opened as well, otherwise it won't connect to the admin port on 5021.

In the end, the only way I could get a client to connect remotely was to pass the hostname of the host to the guest, so that when I connected remotely, the hostname matched what's in DNS.

Here's the command I used to start the server component:

```bash
docker run --rm -it --name=kvlite --hostname=$(hostname) -p 5000:5000 -p 5010-5049:5010-5049 -p 5999:5999 -v $PWD/kvroot:/kvroot oracle/nosql:ce
```

And here's the command I used to connect "remotely" (yes, it's on the same machine, but there's no `--link` and I'm using the same external hostname resolution). 


```bash
docker run --rm -ti oracle/nosql:ce java -jar lib/kvstore.jar runadmin -host $(hostname) -port 5000 -store kvstore
```

Any deviation from this and either the server component doesn't start or the client doesn't connect. 

I haven't tested connectivity via the proxy because the README.md doesn't provide any guidance on how to do that. I'm also wondering if it wouldn't make more sense to have a seperate image for that and use container networking (either Docker's internal networking or Podman CNI) to connect the proxy to the kvlite instance(s).

What is pretty clear is that it doesn't work as documented, so I suggest testing each step again making sure that you remove any existing `/kvroot` content and that you cut and paste what you've documented so that you're testing exactly what a user will run.

Djelibeybi (approved):

NoSQL/README.md
dario-vega: I did in this way to document the difference between the two ranges

- 5010-5020 is the KV_HARANGE
- 5021-5049 is the KV_SERVICERANGE

a user can customize those values, the ranges are not always contiguous. And I realized that docker identifies contiguous ports and merges those. When running `docker ps -a`, I have:

`0.0.0.0:5010-5049->5010-5049/tcp`

I will work with our doc team in order to improve this README documentation

NoSQL/ce/start-kvlite.sh
dario-vega: Thanks for catching this. I will add this to a next version

```
while java -jar lib/kvstore.jar ping -host $HOSTNAME -port 5000  >/dev/null 2>&1 ; [ $? -ne 0 ];do
    echo "Waiting for kvstore to start..."
    sleep 1
done
```

Issue #2266, opened 2022-03-14
oraclelinux7/php/7.4-apache-oracledb breaks after the first update


elmicha:
A container using ghcr.io/oracle/oraclelinux7-php:7.4-apache-oracledb currently breaks the php-oci8 module as so as a `yum, update` is run:

```
========================================================================================================================
 Package                    Arch               Version                            Repository                       Size
========================================================================================================================
Installing:
 php-oci8-21c               x86_64             7.4.28-1.0.3.el7                   ol7_developer_php74             149 k
     replacing  php-oci8-19c.x86_64 7.4.26-1.0.1.el7

...
bash-4.2# php -m
[14-Mar-2022 22:25:08 UTC] PHP Warning:  PHP Startup: Unable to load dynamic library 'oci8.so' (tried: /usr/lib64/php/modules/oci8.so (libclntsh.so.21.1: cannot open shared object file: No such file or directory), /usr/lib64/php/modules/oci8.so.so (/usr/lib64/php/modules/oci8.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
```
I will create a PR that fixes that.

Djelibeybi:
@cjbj can you review this? If the PHP module is automatically updated to 21c, I would've expected the Instant Client to be an available dependency.

PR #2267, opened 2022-03-14
Use php-oci8-21c

elmicha:
As mentioned in https://github.com/oracle/docker-images/issues/2266, before this change the php-oci8 breaks after `yum update`.
With this patch it will continue to work.
See also https://blogs.oracle.com/opal/post/in-place-updates-with-oracle-instant-client-21c-rpms-on-linux, where this oracle-instantclient-release-el7 is listed.

I read https://github.com/oracle/docker-images/blob/53c66772f0d74a7dc102694c331ccec790064e5f/CONTRIBUTING.md and https://oca.opensource.oracle.com/, but frankly I don't know who to ask in my company to sign a Company Agreement.
It's also such a tiny patch it shouldn't even be necessary.


Djelibeybi:
Unfortunately, an OCA is required for all submissions regardless of size.

Djelibeybi:
This also only changes a single image. There are lots of images that use the Instant Client and all of them would need to be updated.

Djelibeybi:
I'm going to close this due to lack of response.

PR #2322, opened 2022-03-29
Support for Containerizing On-premise databases

abhisbyk:
Support for containerizing  on-premise databases by using CLONE_DB option along with README changes.
+ Fixing podman secret create command in the README.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/19.3.0/Dockerfile
+++ b/OracleDatabase/SingleInstance/dockerfiles/19.3.0/Dockerfile
@@ -36,13 +36,14 @@ LABEL "provider"="Oracle"                                               \
 
 # Argument to control removal of components not needed after db software installation
 ARG SLIMMING=true
+ARG INSTALL_FILE_1="LINUX.X64_193000_db_home.zip"
 
 # Environment variables required for this build (do NOT change)
 # -------------------------------------------------------------
 ENV ORACLE_BASE=/opt/oracle \
     ORACLE_HOME=/opt/oracle/product/19c/dbhome_1 \
     INSTALL_DIR=/opt/install \
-    INSTALL_FILE_1="LINUX.X64_193000_db_home.zip" \
+    INSTALL_FILE_1=$INSTALL_FILE_1 \
     INSTALL_RSP="db_inst.rsp" \
     CONFIG_RSP="dbca.rsp.tmpl" \
     PWD_FILE="setPassword.sh" \
--- a/OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile
+++ b/OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile
@@ -36,6 +36,7 @@ LABEL "provider"="Oracle"                                               \
 
 # Argument to control removal of components not needed after db software installation
 ARG SLIMMING=true
+ARG INSTALL_FILE_1="LINUX.X64_213000_db_home.zip"
 
 # Environment variables required for this build (do NOT change)
 # -------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


OracleDatabase/SingleInstance/dockerfiles/19.3.0/Dockerfile
```diff
@@ -36,13 +36,14 @@ LABEL "provider"="Oracle"                                               \
...
 # Argument to control removal of components not needed after db software installation
 ARG SLIMMING=true
+ARG INSTALL_FILE_1="LINUX.X64_193000_db_home.zip"
 
 # Environment variables required for this build (do NOT change)
 # -------------------------------------------------------------
 ENV ORACLE_BASE=/opt/oracle \
     ORACLE_HOME=/opt/oracle/product/19c/dbhome_1 \
     INSTALL_DIR=/opt/install \
-    INSTALL_FILE_1="LINUX.X64_193000_db_home.zip" \
```
gvenzl, line 45: I don't think this line is necessary anymore, will the `ARG INSTALL_FILE_1` be handled like any other environment variable?

OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile
```diff
@@ -36,14 +36,15 @@ LABEL "provider"="Oracle"                                               \
...
 ARG SLIMMING=true
+ARG INSTALL_FILE_1="LINUX.X64_213000_db_home.zip"
 
 # Environment variables required for this build (do NOT change)
 # -------------------------------------------------------------
 ENV ORACLE_BASE=/opt/oracle \
     ORACLE_HOME=/opt/oracle/product/21c/dbhome_1 \
     ORACLE_BASE_HOME=/opt/oracle/homes/OraDB21Home1 \
     INSTALL_DIR=/opt/install \
-    INSTALL_FILE_1="LINUX.X64_213000_db_home.zip" \
```
gvenzl, line 46: I don't think this line is necessary anymore, will the `ARG INSTALL_FILE_1` be handled like any other environment variable?

gvenzl (changes requested):
I believe the `ENV INSTALL_FILE_1=$INSTALL_FILE_1 \` is no longer needed with this change.

OracleDatabase/SingleInstance/dockerfiles/19.3.0/Dockerfile
abhisbyk: In this Dockerfile, we have three stages. The scope of ARG ends with a build-stage ([ref documentation](https://docs.docker.com/engine/reference/builder/#scope)). We will need to replicate this ARG line i.e. `ARG INSTALL_FILE_1="LINUX.X64_193000_db_home.zip"` to each of those build stages. Using ENV mitigate that issue.

OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile
abhisbyk: In this Dockerfile, we have three stages. The scope of ARG ends with a build-stage ([ref documentation](https://docs.docker.com/engine/reference/builder/#scope)). We will need to replicate this ARG line i.e. `ARG INSTALL_FILE_1="LINUX.X64_193000_db_home.zip"` to each of those build stages. Using ENV mitigate that issue.

OracleDatabase/SingleInstance/dockerfiles/19.3.0/Dockerfile
gvenzl: Ah, I see, thanks a lot for the clarification!

OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile
gvenzl: Ah, I see, thanks a lot for the clarification!

gvenzl (approved):

PR #2341, opened 2022-04-12
Add hook to optionally run prestart commands in GoldenGate 21c

tmelhiser:
This allows the container to optionally execute prestart scripts or commands in the container to get it ready before starting the goldengate processes.


OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -121,6 +124,37 @@ function setup_deployment_directories() {
...
+## 
+## The value of ${RUN_COMMAND_BACKGROUND} by default is set to true.
+##     When ${RUN_COMMAND_BACKGROUND} is true:
+##          The command identified by ${RUN_COMMAND} will be run in the background 
+##          and its return status is ignored
+##     When ${RUN_COMMAND_BACKGROUND} is false:
+##          The command identified by ${RUN_COMMAND} will be run in the foreground 
+##          and must return a status of 0 in order to proceed with the start sequence
+##
+function run_prestart_command {
```
sbalousek, line 141: I think this function could consist of a single one-line statement:

```sh
eval "$(RUN_COMMAND}" || abort "${RUN_COMMAND} did not complete successfully!"
```

An empty value for `RUN_COMMAND` becomes a no-op.

If `RUN_COMMAND` is terminated with `&`, the task is executed as a background job. That would also allow for more complex background tasks run with `nohup` and similar wrappers.

sbalousek (changes requested):
Hi @tmelhiser,

I have one comment regarding commands executed in the background (see later), but more importantly, would it make more sense to follow the pattern established for OracleDatabase? That project looks for two directories, `${ORACLE_BASE}/scripts/setup` and `${ORACLE_BASE}/scripts/startup`. The scripts are run before and after the DBMS starts.

If we followed the same pattern, the container for OGG would use a single volume mount for any number of configuration and setup scripts.

- Steve

OracleGoldenGate/21c/bin/deployment-main.sh
tmelhiser: The mounted volume pattern is nice. I think this is the route I will approach as it gives a known set of tested scripts that are run.

I was not a great fan of the arbitrary RUN_COMMAND approach to begin with, due to the ability to run any arbitrary code in the container as a privileged user.  I avoided the eval statement even more so, due to it's expansion allowing for even more complex execution patterns.

The directory/mounted volume approach appears to force a foreground linear lexicographical approach.

Is there any desire to allow for asynchronous background processes to be started as well?

This could be achieved with an intuitive directory structure:

| - hooks
&nbsp;&nbsp;&nbsp;&nbsp;| -- background
&nbsp;&nbsp;&nbsp;&nbsp;| -- prestart

Although possible, I did not propose an exit/cleanup folder.  This enhancement could be augmented in the future, but the current signal handling and health monitoring would need a good bit of rework to ensure OGG was in the correct state while not having the container terminated/restarted by the container management system prior to the scripts finishing.

Additional feedback would be appreciate before proceeding.

Thanks again.

tmelhiser: Updated to use the directory pattern ${OGG_HOME}/scripts/setup and ${OGG_HOME}/scripts/startup

The backgrounding issue can be handled on a per script basis inside the script itself, rather than as an arbitrary flag in the config.  Allowing user determination on what runs in the background and in what order.

The scripts will be run lexicographically and will recursively span subdirectories.
The order preference of files is determined only by the lexicographical order on a per directory basis.
ie: 
> 02-sanity.sh
> 01-connectivity-validation\
>>> 02-test-hostA.sh 
>>>01-test-hostZ.sh 

will be run in the lexicographic order:
1. 01-test-hostZ.sh
2. 02-test-hostA.sh
3. 02-sanity.sh

SQLPlus is dynamically located and .sql files can be run.
Scripts in */setup will be run prior to any other startup sequence
Scripts in */startup will be run after ogg/nginx successfully start

OracleGoldenGate/21c/Dockerfile
```diff
@@ -14,7 +14,8 @@ COPY          install-*.sh      /tmp/
 COPY          ${INSTALLER}      /tmp/installer.zip
 COPY          bin/              /usr/local/bin/
 
-RUN           bash -c  /tmp/install-prerequisites.sh && \
+RUN           chmod 755 /usr/local/bin/*             && \
```
sbalousek, line 17: This should be handled by the file mode stored in Git. Did you run into problems on Windows maybe?

OracleGoldenGate/21c/README.md
```diff
@@ -66,8 +66,9 @@ $ docker run \
     -e OGG_ADMIN=<admin user name> \
     -e OGG_ADMIN_PWD=<admin password> \
     -e OGG_DEPLOYMENT=<deployment name> \
-    -e RUN_COMMAND=<script or command> \
-    -e RUN_COMMAND_BACKGROUND=<true | false> \
+    -e SETUP_USER_SCRIPTS=<path to setup scripts> \
```
sbalousek, line 69: The convention established in the `OracleDatabase` project is a pre-coded location of either:
- `"$ORACLE_BASE"/scripts/setup`
- `"$ORACLE_BASE"/scripts/startup`

I would expect something similar here, with:
- `"$OGG_HOME"/scripts/setup`
- `"$OGG_HOME"/scripts/startup`

Is there really any value in allowing control over the script directory location when a volume mount already controls the content of those directories?

OracleGoldenGate/21c/README.md
```diff
@@ -66,8 +66,9 @@ $ docker run \
...
+    -e STARTUP_USER_SCRIPTS=<path to startup scripts> \
+    -e ABORT_ON_USER_SCRIPT_ERRORS=<true | false>
```
sbalousek, line 71: It might be better to handle errors from within the scripts. Perhaps if there was a generic `abortContainer.sh` command that terminated the container with an error message called from the user script?

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -30,6 +27,14 @@ function abort() {
 :     "${OGG_HOME:?}"
 [[ -d "${OGG_HOME}"            ]] || abort "Deployment runtime, '${OGG_HOME}'. not found."
 
+:     "${ABORT_ON_USER_SCRIPT_ERRORS:=true}"
+:     "${SETUP_USER_SCRIPTS:=${OGG_HOME}/scripts/setup}"
+[[ -d "${SETUP_USER_SCRIPTS}" ]] || mkdir -p "${SETUP_USER_SCRIPTS}"
```
sbalousek, line 32: These `mkdir` commands can go in the `Dockerfile`

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -30,6 +27,14 @@ function abort() {
...
+:     "${STARTUP_USER_SCRIPTS:=${OGG_HOME}/scripts/startup}"
+[[ -d "${STARTUP_USER_SCRIPTS}" ]] || mkdir -p "${STARTUP_USER_SCRIPTS}"
+
+SQLPLUS="$(find ${OGG_HOME} -name sqlplus)"
```
sbalousek, line 36: Since OGG is not the DBMS, any SQL processing should be handled entirely in the user scripts.

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -125,34 +130,69 @@ function setup_deployment_directories() {
...
+                    run_user_scripts "${f}"
+                } || {
+                    case "$f" in
+                        *.sh)
+                                printf "%s: running %s\n" "${0}" "${f}";
+                                . "$f" && state=$? || state=$?
+                                ;;
+                        *.sql)
+                                printf "%s: running %s\n" "${0}" "${f}";
+                                echo "exit" | LD_LIBRARY_PATH=$(dirname ${SQLPLUS}):${LD_LIBRARY_PATH} ${SQLPLUS} -s "/ as sysdba" @"$f" && state=$? || state=$?
```
sbalousek, line 174: We can't execute SQL*Plus from here because the database is likely not in the same container. When running in another container, credentials are needed, etc.

Because of the inherent complexities, I would like to see support for `*.sql' files removed.

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -125,34 +130,69 @@ function setup_deployment_directories() {
...
+    } || {
+        if [ -d "$SCRIPTS_ROOT" ] && [ -n "$(ls -A "$SCRIPTS_ROOT")" ]; then
+            printf "Executing user defined scripts in: %s\n" "${SCRIPTS_ROOT}"
+        
+            for f in "${SCRIPTS_ROOT}"/*; do
+                [ -d "${f}" ] && {
+                    run_user_scripts "${f}"
+                } || {
+                    case "$f" in
+                        *.sh)
```
sbalousek, line 168: I would vote for executing anything with the executable bit set. That way, Python scripts (or whatever) could also be run. But, the Database only supports `*.sh` files, so maybe we should stick to just those.

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -125,34 +130,69 @@ function setup_deployment_directories() {
...
+                                printf "%s: ignoring %s\n" "${0}" "${f}" && state=$? || state=$?
+                                ;;
+                    esac
+                    [ $state -ne 0 ] && {
+                        printf "#################\n"
+                        printf "## WARNING - %s: user script failed [exit code: %s]: %s\n" "${0}" "${state}" "${f}"
+                        printf "#################\n"
+                        [ "${ABORT_ON_USER_SCRIPT_ERRORS,,}" == "true" ] && {
+                            printf "Aborting startup! - Sleeping 10 seconds before exit\n"
+                            sleep 10;
```
sbalousek, line 186: Sleep?  I no sleep!
- Gollum

What are you thinking here with the pause?

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -182,11 +222,16 @@ function start_nginx() {
 ##  Termination handler
 ##
 function termination_handler() {
-    [[ -z "${ogg_pid}" ]] || {
-        kill  "${ogg_pid}"
-        unset    ogg_pid
+    [[ -f "${OGG_HOME}/ogg.pid" ]] && {
+        kill  $(cat "${OGG_HOME}/ogg.pid")
```
sbalousek, line 226: Er. Why kill the Service Manager here?

OracleGoldenGate/21c/install-deployment.sh
```diff
@@ -82,6 +82,11 @@ EOF
     else
         $(run_as_ogg) tar xf /tmp/installer/*.tar -C "${OGG_HOME}"
     fi
+
```
sbalousek, line 85: These should be in `Dockerfile`, and any permissions mismatches between the mounted scripts volume will need to be addressed in the host environment.

sbalousek (changes requested):
Hi @tmelhiser,

I like the direction this is heading in. I have quite a few questions and one major correction with regard to the script directory location.

- Steve

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -125,34 +130,69 @@ function setup_deployment_directories() {
...
+            printf "Executing user defined scripts in: %s\n" "${SCRIPTS_ROOT}"
+        
+            for f in "${SCRIPTS_ROOT}"/*; do
+                [ -d "${f}" ] && {
+                    run_user_scripts "${f}"
+                } || {
+                    case "$f" in
+                        *.sh)
+                                printf "%s: running %s\n" "${0}" "${f}";
+                                . "$f" && state=$? || state=$?
```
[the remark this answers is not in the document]
tmelhiser: I like the idea of sourcing the file, since many times values derived from one script can be reused/passed to subsequent ones.

OracleGoldenGate/21c/bin/deployment-main.sh
tmelhiser: I really like the idea of checking the execute bit.  The issue I worry about there is the oddities we see in git check ins where those aren't stored correctly.

I will defer to your call here.

OracleGoldenGate/21c/bin/deployment-main.sh
tmelhiser: I will drop off the sqlplus hook.  I assume this also means dropping the script execution of *.sql files?

OracleGoldenGate/21c/bin/deployment-main.sh
tmelhiser: This kind of goes back to the question of "Do we allow user defined locations of where the *_USER_SCRIPTS are located".  If we allow user definition, this is a nice safeguard preventing errors in the belly of the script due to missing dirs.

OracleGoldenGate/21c/bin/deployment-main.sh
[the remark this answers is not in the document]
sbalousek: Yes, but:

- The `source` command can be used from those scripts.
- Using `source` here potentially corrupts the execution environment

Since the `source` option is used by the DBMS team, we should use it here, too.

OracleGoldenGate/21c/bin/deployment-main.sh
sbalousek: Let's stick to the DBMS model and only process `*.sh` files via `source` (or `.`)

OracleGoldenGate/21c/bin/deployment-main.sh
sbalousek: Since the DBMS containerized image does not allow a user-defined location for scripts, I would like to use a fixed location here, too.

OracleGoldenGate/21c/README.md
tmelhiser: If I understand correctly, the abortContainer.sh would be expected to be called by the third party scripts as part of their error handling?  ie: their trap conditions would have to implement handing control to abortContainer.sh, which would then have to handle determining whether to abort or hand control back to the originating deployment-main.sh or continue with the existing thrid party script?

This seems like a pretty complex scripting model for a large majority of the user base to comprehend.

If simplification and avoidance confusion by the ABORT_ON_USER_SCRIPT_ERROR is the goal,  I would suggest the same as you -- any script non-zero exit aborts the boot; forcing the users to handle errors in their scripts.

OracleGoldenGate/21c/bin/deployment-main.sh
tmelhiser: The OGG primary instance never receives the os level termination signal.  The original code trying to trap the OGG PID is actually only trapping the PID of the tail command.

Since the PID of the OGG instance is created inside the python script, it must be gathered there.  In order to issue the shutdown of OGG when the main PID 1 is signaled to exit by the orchestration layer (docker/k8s/etc) the trap then kills the child processes.  The PID of the OGG instance was never really being notified.

This correctly sends the exit signal to the OGG process and it shutsdown cleanly, rather than waiting the container timeout period and forcefully killing all remaining PIDs.

OracleGoldenGate/21c/install-deployment.sh
tmelhiser: I don't think I changed anything in this area other than formatting?  The extra space between the -p and the ${OGG_HOME}.

Probably just me being pedantic about the spacing... if you prefer the space, I can put it back.

OracleGoldenGate/21c/Dockerfile
tmelhiser: No, this was built on a linux machine.

sbalousek: OK. Well, the `install-prerequisites.sh` script is required. I believe the changes to `Dockerfile` should be reverted,

OracleGoldenGate/21c/bin/deployment-main.sh
sbalousek: In Docker, the logs for a container are available until the container is destroyed.

Regardless, if it's specific to your environment, it should be in a user script. That would follow the behavior in thee DBMS container.

OracleGoldenGate/21c/install-deployment.sh
sbalousek: Unless reformatting with a purpose, whitespace-only changes should be avoided.

OracleGoldenGate/21c/bin/deployment-main.sh
sbalousek: I'm not sure what you mean by 'OGG primary instance.'  As far as I know, all processes in a Docker container are sent the SIGTERM signal when the container is terminated. Service Manager, in particular, shuts down pretty quickly when it receives SIGTERM.

I would like to better understand which processes are not responding in a timely fashion to SIGTERM.

Anyway, this change is not really part of enable user scripts, so it should be in a separate PR at a minumum.

OracleGoldenGate/21c/README.md
sbalousek: The `abortContainer.sh` script was just a thought and not something I expected to be part of this PR. IN the DBMS scripts (see `OracleDatabase/SingleInstance/dockerfiles/21.3.0/runUserScripts.sh`), there is no error handling.

Although I am not fond of that naive approach, I believe that whatever is implemented for OGG should be as similar as possible to what is already being done for the DBMS images. The reason for that belief is consistent implementation and behavior.

abhijitvk (commented):
@sbalousek : All the requested changes are completed and are ready for your re-review. Would appreciate your time to get this reviewed and provide feedback and /or merge at the earliest convenience.

OracleGoldenGate/21c/install-deployment.sh
abhijitvk: reverted the whitespace changes

abhijitvk:
@sbalousek Requesting feedback on the PR and merge if all the earlier comments have been addressed. Thank you!

OracleGoldenGate/21c/Dockerfile
sbalousek: The use of `chmod` is not required. All shell scripts and other executables already have the executable bit set in this Git repository. These changes can be reverted.

OracleGoldenGate/21c/bin/deployment-init.py
```diff
@@ -133,11 +133,17 @@ def get_digest_authentication():
     userName, credential = get_admin_credentials()
     return requests.auth.HTTPDigestAuth(userName, credential)
 
+def get_basic_auth_authentication():
```
sbalousek, line 136: Although a nice addition, these changes are unrelated to the focus of this PR, which is startup scripts. Please open a separate issue regarding Basic/Digest authentication and create a PR for that issue separately,

OracleGoldenGate/21c/README.md
sbalousek: This issue needs resolution.

sbalousek (changes requested):

sbalousek (changes requested):
Style changes are not desirable. Please remove.

sbalousek (changes requested):
These changes have nothing to do with the issue of running scripts at Deployment startup.

OracleGoldenGate/21c/bin/deployment-init.py
abhijitvk: While trying to run the container each time I was seeing the error stacktrace as below on the Linux machine. Basic auth is supported by the API calls and should work absolutely fine. Fallback would be the digest authentication. I will create a separate PR for this one if that makes more sense.

`2022-06-01T13:20:23.070Z | INFO    | ALL DONE (0 errors).
Traceback (most recent call last):
  File "/usr/local/bin/deployment-init.py", line 265, in <module>
    sys.exit(main())
  File "/usr/local/bin/deployment-init.py", line 260, in main
    create_ogg_deployment()
  File "/usr/local/bin/deployment-init.py", line 251, in create_ogg_deployment
    reset_service_configuration(serviceName)
  File "/usr/local/bin/deployment-init.py", line 166, in reset_service_configuration
    config = get_service_config(serviceName)
  File "/usr/local/bin/deployment-init.py", line 140, in get_service_config
    response = get_requests_session().get(url, headers=rest_call_headers, auth=get_digest_authentication())
  File "/usr/lib/python3.9/site-packages/requests/sessions.py", line 555, in get
    return self.request('GET', url, **kwargs)
  File "/usr/lib/python3.9/site-packages/requests/sessions.py", line 542, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3.9/site-packages/requests/sessions.py", line 662, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/usr/lib/python3.9/site-packages/requests/hooks.py", line 31, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/usr/lib/python3.9/site-packages/requests/auth.py", line 267, in handle_401
    prep.headers['Authorization'] = self.build_digest_header(
  File "/usr/lib/python3.9/site-packages/requests/auth.py", line 185, in build_digest_header
    HA1 = hash_utf8(A1)
  File "/usr/lib/python3.9/site-packages/requests/auth.py", line 148, in md5_utf8
    return hashlib.md5(x).hexdigest()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS `

OracleGoldenGate/21c/Dockerfile
abhijitvk: I took the chmod out and if needed can be done via Dockerfile add-on or user script

OracleGoldenGate/21c/bin/deployment-main.sh
abhijitvk: replaced with the source for processing *.sh files

OracleGoldenGate/21c/README.md
abhijitvk: The issue has been resolved now. Not allowing any more control over the scripts directory as the volume mount controls it.

OracleGoldenGate/21c/bin/deployment-main.sh
abhijitvk: Updated to be the fixed location

abhijitvk:
@sbalousek Feedback changes are all completed and ready for review + merge. Appreciate your quick response

OracleGoldenGate/21c/README.md
```diff
@@ -141,6 +143,15 @@ Oracle GoldenGate Administration Client for Oracle
 Version 21.3.0.0.0 ...
 ```
 
+### Running scripts before setup and on startup
```
abhijitvk, line 146: @sbalousek  Adding this new section in README for more clarity

abhijitvk:
Are these changes ready to be merged? @sbalousek

OracleGoldenGate/21c/README.md
```diff
@@ -141,6 +143,15 @@ Oracle GoldenGate Administration Client for Oracle
...
+The container images can be configured to run scripts before setup and on startup. Currently, sh extensions are supported. For setup scripts just mount the volume /u01/ogg/scripts/setup or extend the image to include scripts in this directory. For startup scripts just mount the volume /u01/ogg/scripts/startup or extend the image to include scripts in this directory. Both of those locations are static and the content is controlled by the volume mount.
```
sbalousek, line 147: Please back-tick quote literals like `.sh` and directory names

OracleGoldenGate/21c/README.md
```diff
@@ -141,6 +143,15 @@ Oracle GoldenGate Administration Client for Oracle
...
+
+The example below mounts the local directory myScripts to /u01/ogg/scripts which is then searched for custom startup scripts:
```
sbalousek, line 149: Backtick quote, please

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -121,6 +126,53 @@ function setup_deployment_directories() {
...
+##
+## Scripts are run lexicographically and recursively from the directories pointed to by:
+##      - ${SETUP_USER_SCRIPTS} are executed prior to any other steps in the boot sequence
+##      - ${STARTUP_USER_SCRIPTS} are executed after ogg/nginx startup
+##
+function run_user_scripts {
+
+    local SCRIPTS_ROOT="${1}";
+
+    [ -z "$SCRIPTS_ROOT" ] && {
```
sbalousek, line 145: `${SCRIPTS_ROOT}` will never be empty

OracleGoldenGate/21c/install-deployment.sh
```diff
@@ -82,6 +82,11 @@ EOF
...
+    for dir in "${OGG_HOME}/scripts/setup" "${OGG_HOME}/scripts/startup"; do
+        mkdir -p "${dir}"
+        chown -R ogg:ogg "$(dirname "${dir}")"
+    done
```
sbalousek, line 89: or:
```.sh
mkdir -p '${OGG_HOME}/scipts/"{setup,startup}
chown -R ogg:ogg "${OGG_HOME}/scripts"
```

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -153,6 +205,7 @@ function termination_handler() {
         unset    ogg_pid
     }
     [[ ! -f "/var/run/nginx.pid" ]] || {
+        echo "stopping nginx"
```
sbalousek, line 208: Is this really needed?

OracleGoldenGate/21c/bin/deployment-main.sh
```diff
@@ -27,6 +27,11 @@ function abort() {
 :     "${OGG_HOME:?}"
 [[ -d "${OGG_HOME}"            ]] || abort "Deployment runtime, '${OGG_HOME}'. not found."
 
+:     "${SETUP_USER_SCRIPTS=${OGG_HOME}/scripts/setup}"
```
sbalousek, line 30: Can these two variables be reduced to simply `OGG_DEPLOYMENT_SCRIPTS`?

sbalousek (changes requested):
I feel like these changes are very close. In addition, I believe the `Dockerfile` should include the environment variables for the scripts directory and export that directory as a volume.

OracleGoldenGate/21c/bin/deployment-main.sh
abhijitvk: Combined into a single variable `OGG_DEPLOYMENT_SCRIPTS`

OracleGoldenGate/21c/install-deployment.sh
abhijitvk: replaced for loop with suggested mkdir single execution

OracleGoldenGate/21c/README.md
abhijitvk: added Backtick quote

OracleGoldenGate/21c/README.md
abhijitvk: added Backtick quote

OracleGoldenGate/21c/bin/deployment-main.sh
abhijitvk: It won't be assuming that the function `run_user_scripts` is invoked with a non-empty value. Additional check should be OK i guess.

abhijitvk:
All the suggested changes are completed. Should be ready for merge at this time. let me know for any other feedback

abhijitvk:
> Thank you for all your efforts on this. I hope all my change requests did not drive you crazy!

Sure thing! Thanks for your valuable feedback. Is there a process before merging into the main branch?

PR #2358, opened 2022-05-03
April cpu patch

aureliogrb:
Updated the versions to match the May patch update that replaced the April 2022 CPU.  Also changed the dockerfiles for JDK 11, 17, and 18 to be able to build either x86-64 or aarch64. (serverJRE 8 is only available as x86-64 that version is just an update from 8u331 to 8u33)

### DIFF
--- a/OracleJava/18/build.sh
+++ b/OracleJava/18/build.sh
@@ -5,4 +5,4 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 echo "Building Oracle JDK 18 on Oracle Linux 8"
-docker build --platform amd64 --file Dockerfile --tag oracle/jdk:18  --tag oracle/jdk:18-oraclelinux8 .
+docker build  --file Dockerfile --tag oracle/jdk:18  --tag oracle/jdk:18-oraclelinux8 .
--- a/OracleJava/18/build.sh
+++ b/OracleJava/18/build.sh
@@ -5,4 +5,4 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 echo "Building Oracle JDK 18 on Oracle Linux 8"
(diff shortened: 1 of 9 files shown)


Djelibeybi:
Please check the diff output as you have added a few extra spaces that should be removed.

OracleJava/18/build.sh
```diff
@@ -5,4 +5,4 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 echo "Building Oracle JDK 18 on Oracle Linux 8"
-docker build --file Dockerfile --tag oracle/jdk:18  --tag oracle/jdk:18-oraclelinux8 .
\ No newline at end of file
+docker build  --file Dockerfile --tag oracle/jdk:18  --tag oracle/jdk:18-oraclelinux8 .
```
rjeberhard, line 8: Unnecessary space added

OracleOpenJDK/18/Dockerfile
```diff
@@ -21,11 +21,11 @@
 #		$ bash build.sh
 #
 
-FROM oraclelinux:8
+FROM oraclelinux:8 
 
 MAINTAINER Aurelio Garcia-Ribeyro <aurelio.garciaribeyro@oracle.com>
 
-ENV JAVA_PKG=https://download.java.net/java/GA/jdk18.0.1/3f48cabb83014f9fab465e280ccf630b/10/GPL/openjdk-18.0.1_linux-x64_bin.tar.gz \
+ENV JAVA_URL=https://download.java.net/java/GA/jdk18.0.1.1/65ae32619e2f40f3a9af3af1851d6e19/2/GPL \
```
rjeberhard, line 28: Just curious... Does download.java.net not also have a "latest" URL as download.oracle.com does?

aureliogrb: download.java.net does not have the "latest" option as download.oracle.com does.  It is something that we have considered and might still add in the future.   If that happens, I will update the scripts accordingly

Djelibeybi (approved):

PR #2372, opened 2022-05-27
fix isOGGRunning to avoid conflict with tailReport ../ServiceManager.log

sscobici:
In exec_microservices function I commented startReverseProxy line and tailReport command is executed in parallel with isOGGRunning.
Because tailReport is listening to ServiceManager.log isOGGRunning thinks that ServiceManager is already running and doesn't start it.

function exec_microservices {
    export OGG_ETC_HOME="${OGG_DEPLOY_BASE}/ServiceManager/etc"
    export OGG_VAR_HOME="${OGG_DEPLOY_BASE}/ServiceManager/var"
    isOGGRunning || ${runAsUser} ${OGG_HOME}/bin/ServiceManager &
    # startReverseProxy
    tailReport "${OGG_VAR_HOME}/log/ServiceManager.log"
}

### DIFF
--- a/OracleGoldenGate/12c-19c/runOracleGoldenGate.sh
+++ b/OracleGoldenGate/12c-19c/runOracleGoldenGate.sh
@@ -29,7 +29,7 @@ function getCommand {
 
 getCommand runuser
 runAsUser="${RUNUSER} -u oracle --"
-OGGProcesses="(adminclient|adminsrvr|distsrvr|extract|ggsci|pmsrvr|recvsrvr|replicat|server|ServiceManager)"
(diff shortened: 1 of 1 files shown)


sbalousek (changes requested):
I don't see how commenting out `startReverseProxy` has any bearing on things. Moreover, `tailReport` is run *after* `isOGGRunning`, and not in parallel.

I agree there is a bug in `isOGGRunning`, but the correct fix is to change `-f` to `-x` on line 115.

sbalousek:
Closing request due to lack of response.

@sscobici - Please create a new issue describing the problem you are trying to solve here.

PR #2405, opened 2022-06-29
use default expansions for run as agent user variable

ChaitanyaKulkarni28:
set -u in a script causes the script to abort if the variable being referenced is not set, using default expansions for the RUN_AGENT_AS_USER variable


niravg75 (approved):

ChaitanyaKulkarni28:
> Please remember to sign off your commits using `--signoff` or `-s` when running `git commit` in future.

Yes, thanks. Completely missed that

PR #2408, opened 2022-07-07
Add OCI CLI Docker Image

amcheste:
This PR provides a dockerfile that can be used to build an image containing the OCI CLI.

This will allow users to run the OCI CLI on their laptop without installing any dependencies or from a container orchestrator such as Kubernetes.

Example Usage:
```
$ alias cli="docker run -it -v $HOME/.oci:/home/oracle/.oci docker.io/oracle/oci-cli:latest oci"
$ oci os ns get
```

### DIFF
--- a/OracleCloudInfrastructure/oci-cli/Dockerfile
+++ b/OracleCloudInfrastructure/oci-cli/Dockerfile
@@ -1,0 +1,12 @@
+FROM oraclelinux:8
+
+RUN dnf update -y && \
+    dnf install -y oraclelinux-developer-release-el8 && \
+    dnf install -y python36-oci-cli
+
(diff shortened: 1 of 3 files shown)


OracleCloudInfrastructure/oci-cli/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM oraclelinux:8
```
Djelibeybi, line 1:
```suggestion
FROM ghcr.io/oracle/oraclelinux8-python:3.6
```

This is to ensure the image is updated automatically by our CI/CD system to use the latest Oracle Linux and Python packages.

Djelibeybi:
Closing after internal discussions.

Issue #2413, opened 2022-07-15
Missing Documentation for deletion of node from RAC Custer


ahujasa:
https://github.com/oracle/docker-images/tree/main/OracleDatabase/RAC/OracleRealApplicationClusters is missing Documentation for deletion of node from RAC Custer

ahujasa:
Found way to do though, pls add to documentation-

docker exec racnode1 /bin/sh -c "sudo -E /opt/scripts/startup/DelNode.sh <node_to_be_deleted>"
docker stop  <<<node_to_be_deleted>>>
docker rm <<node_to_be_deleted>>>

ahujasa:
Raised merge request https://github.com/oracle/docker-images/pull/2416

PR #2416, opened 2022-07-18
Document delete of racnode

saurabhaahujaa:
Document delete of racnode
Fix for https://github.com/oracle/docker-images/issues/2413


ahujasa:
BTW, i also noticed that after deleting node as above, it doesnt delete entry from hostfile, its has to be deleted manually.

PR #2422, opened 2022-07-22
how to fix a connection issue from oracle container

fabriciogl:
Added a line to explain how to fix a common issue of connection to oracle container as it's reported on Stackoverflow (https://stackoverflow.com/questions/19660336/how-to-approach-a-got-minus-one-from-a-read-call-error-when-connecting-to-an-a)

Third answer.

### DIFF
--- a/OracleDatabase/SingleInstance/README.md
+++ b/OracleDatabase/SingleInstance/README.md
@@ -135,6 +135,8 @@ Once the container has been started and the database created you can connect to 
     sqlplus system/<your password>@//localhost:1521/<your SID>
     sqlplus pdbadmin/<your password>@//localhost:1521/<Your PDB name>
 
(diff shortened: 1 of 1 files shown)


cjbj:
Shouldn't this be added to https://github.com/oracle/docker-images/blob/main/OracleDatabase/SingleInstance/FAQ.md instead?

fabriciogl:
Maybe, but the official instruction to use `localhost` don't actually works. So.. idk if worth to keep at first page instructions that doesn't work and put in a separate place instruction that really works.

aditya059 (approved):

aditya059:
Hi @fabriciogl 

I have not faced this connection issue. We think this PR (#2607) may fix the issue. Can you please try once this PR is merged and let us know whether you still face this connection issue. For now, we are approving your PR.

Thanks

PR #2431, opened 2022-08-17
Oracle Analytics Server 6.4 public dockerfiles

padambengani:


padambengani:
Note that PM Alan Lee (alan.w.lee@oracle.com) has reviewed this.

Djelibeybi (changes requested):
Please replace all generic uses of the word "Docker" in the `README.md` with "container" instead, i.e. "Sample container configurations" not "Sample Docker configurations". The only time you should refer to Docker specifically is when providing an example that uses the `docker` command.

Also, why do your two `Dockerfiles` use a different base image? The one inside `dockerfiles/6.4` uses `oracle/fmw-infrastructure` while the one inside `patches/6.4-patch` uses `oracle/biplatform`. 

Also, your copyright notices are incorrect. They should use `2022` as the year, not `2017`, they should not include the phrase `All rights reserved.` and they should use `https` for the URL to the Universal Permissive License. Please fix all of them.

Djelibeybi:
@mriccell @rjeberhard when are we going to get `fmw-infrastructure` images that are based on more recent versions of Oracle Linux and Oracle Java? I'm not sure we should be releasing new container images based on Oracle Linux 7 now, as it has less than two years left of premier support.

rjeberhard:
> @mriccell @rjeberhard when are we going to get `fmw-infrastructure` images that are based on more recent versions of Oracle Linux and Oracle Java? I'm not sure we should be releasing new container images based on Oracle Linux 7 now, as it has less than two years left of premier support.

@Djelibeybi, I'm certainly not opposed to releasing images on Oracle Linux 9. I don't know about @mriccell's experience, however, but I've not yet had a customer ask for Oracle Linux 9 based images or to complain about Oracle Linux 7 -- let me see if I can find the statistics about what has been pulled to see if customers are moving to Oracle Linux 8.

Oracle Java is a different story... WebLogic 12.2.1.3 and 12.2.1.4 only run on JDK 8. WebLogic 14.1.1.0 runs on JDK 8 and JDK 11. We (the WebLogic team) are actively working on support for more recent JDK's but this is a massive undertaking. Even when versions support newer JDK releases it will likely be a long, long time before a critical mass of customers upgrade from the versions above. 

Therefore, we have instead made sure to negotiate with the Oracle Java team extended support lifetimes. You will notice that our images are frequently published with updated JDK patch levels.

padambengani:
> Please replace all generic uses of the word "Docker" in the `README.md` with "container" instead, i.e. "Sample container configurations" not "Sample Docker configurations". The only time you should refer to Docker specifically is when providing an example that uses the `docker` command.
> 
> Also, why do your two `Dockerfiles` use a different base image? The one inside `dockerfiles/6.4` uses `oracle/fmw-infrastructure` while the one inside `patches/6.4-patch` uses `oracle/biplatform`.
> 
> Also, your copyright notices are incorrect. They should use `2022` as the year, not `2017`, they should not include the phrase `All rights reserved.` and they should use `https` for the URL to the Universal Permissive License. Please fix all of them.

@Djelibeybi I have fixed the copyright / licensing info as per your comments and also replaced various occurrences of word Docker with Container in README file. Please confirm if the docker needs to be replaced in the following lines as well - 

Oracle Analytics Server on Docker
## Oracle Analytics Server Docker Image Creation

Regarding your question about two dockerfiles using different base images, its because the one inside "patches/6.4-patch" is for building OAS 6.4 patched image and it is supposed to be run after OAS 6.4 base image is built (i.e. dockerfiles/6.4). I have added the sections in the README files that this OAS 6.4 patched image needs to be built after OAS 6.4 base image is built.

Djelibeybi:
Yes, all references to Docker must be removed unless you're specifically demonstrating the use of the `docker` command-line utility. These images are designed to run your product in a container that may or may not be powered by Docker.

Djelibeybi:
My confusion about the patched image comes from the new name being Oracle Analytics Server and references to OAS while the image is actually created as `biplatform` and then patched. It would be good to call this out much more explicitly and perhaps use `oracle/analyticsserver` as your image name instead of `biplatform` to reduce confusion.

Djelibeybi (changes requested):
Unfortunately, using find and replace to replace "Docker" with "Container" is not what I meant. You need to contextually replace the use of the word Docker with either a generic reference to containers or by removing it completely.

For example, the title should become "Oracle Analytics Server for Containers" or "Containerised Oracle Analytics Server". Likewise, instead of "on Docker" (which was incorrect itself) it should become "in a container".

As Oracle considers the content of README.md files to be official documentation, I strongly encourage you to engage your doc writing team to review and edit the content.

padambengani:
@Djelibeybi Doc team has updated the README, please have a look.

Djelibeybi:
Folks, please ensure that all your accounts have been added to the Oracle GitHub organization and that your membership is public so that our OCA bot doesn't think you're external.

Djelibeybi:
I have marked this PR as a draft until the commit ownership is rectified. If necessary, close this PR and open a new one with a clean set of commits that only contain Oracle email addresses from users with publicly verifiable membership to the Oracle GitHub Organization.

If you need assistance with any of this, please contact me on Slack.

Djelibeybi:
Lastly, please use [`markdownlint`](https://github.com/markdownlint/markdownlint) to lint all Markdown files so that a consistent, GitHub-flavoured Markdown style is used throughout. All shell scripts should be checked using [`shellcheck`](https://www.shellcheck.net/).

Both of these are available as Visual Studio Code plugins and other forms, so find the variant that works best for you.

Djelibeybi:
I am closing this PR due to lack of activity. Please resubmit a new PR with all commits from users who are members of OGHO. Also, please double-check the copyright headers as I noticed somew incorrect formatting and dates in this PR.

PR #2435, opened 2022-08-19
- Install JDK 11 during setup instead of JDK 8.

amar143105:
my docker

### DIFF
--- a/OracleGoldenGate/21c/install-prerequisites.sh
+++ b/OracleGoldenGate/21c/install-prerequisites.sh
@@ -7,7 +7,7 @@ set -e
 ##  Installation of prerequisite software
 ##
 
-packages=(java-1.8.0-openjdk jq libaio libnsl python39-requests python39-psutil tar unzip xz)
+packages=(java-11-openjdk jq libaio libnsl python39-requests python39-psutil tar unzip xz)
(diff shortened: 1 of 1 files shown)


OracleGoldenGate/21c/install-prerequisites.sh
```diff
@@ -7,7 +7,7 @@ set -e
 ##  Installation of prerequisite software
 ##
 
-packages=(java-1.8.0-openjdk jq libaio libnsl python39-requests python39-psutil tar unzip xz)
+packages=(java-11-openjdk jq libaio libnsl python39-requests python39-psutil tar unzip xz)
```
sbalousek, line 10: I'll need to check if JDK11 is supported

sbalousek: @amar143105 - Please explain your intent behind this change because I don't think it accomplishes anything of substance. Specifically, the JDK distributed with OGG in `/u01/ogg/jdk` will be the JDK used (for OGG Big Data) and not the JDK installed here. The JDK installed to the system from this file is only used in `deployment-init.py`

sbalousek:
Rejecting pull request with no response.

PR #2457, opened 2022-09-13
Expose OCI CLI image build time values at runtime via environment variables

Djelibeybi:
Also removed the hard-coded repo owner string to allow downstream forks to successfully test/run the workflow.


Djelibeybi:
@AmedeeBulle could you please review?

PR #2467, opened 2022-09-21
Remove the compat image from the build workflow

Djelibeybi:
I have no idea if this is correct, I just removed all mention of the compat image. If it's not correct, feel free to close this PR and submit another one.


AmedeeBulle (changes requested):
I spotted a small issue, the rest looks OK, but I'll do a second pass tomorrow with a fresh mind...

.github/workflows/build-and-push-dev-images.yml
```diff
@@ -78,30 +74,14 @@ jobs:
               echo "PUSH: Rebuilding changed images only"
```
AmedeeBulle, line 74: There is still a `compat=0` / `compat=1` we can remove a few lines above this one

AmedeeBulle (approved):
LGTM.

There are still 2 lines setting the the unused `compat` variable -- we can remove these now or address in a separate PR, it doesn't hurt.

PR #2509, opened 2022-11-25
Updated to support podman 4.2.0

rishiagarwal-oracle:
Signed-off-by: Rishi Agarwal <rishi.agarwal@oracle.com>

Minor change in LCM script to make it compatible with latest podman versions (4.2.0 +)

### DIFF
--- a/OracleIdentityGovernance/samples/scripts/agentManagement.sh
+++ b/OracleIdentityGovernance/samples/scripts/agentManagement.sh
@@ -153,7 +153,7 @@ detectContainerRuntime()
 
   if [ -z "$containerRuntime" -a "$containerRuntime" == "" ]
   then
-    echo "ERROR: No container runtime available. Please install docker/podman before proceeding"
+    echo "ERROR: No container runtime available. Please install Docker/Podman before proceeding"
     errorFlag=true
     return
   fi
--- a/OracleIdentityGovernance/samples/scripts/agentManagement.sh
+++ b/OracleIdentityGovernance/samples/scripts/agentManagement.sh
@@ -226,14 +226,10 @@ loadImage(){
   imageName=""
   if [ "$containerRuntime" == "docker" ]
   then
-    imageName=$(docker load < "$AGENTDIR"/agent-lcm/idm-agcs-agent-framework.dockerize_agent.tar.gz | grep "Loaded image:" | awk '{ print $3 }')
+    imageName=$(docker load < "$AGENTDIR"/agent-lcm/idm-agcs-agent-framework.dockerize_agent.tar.gz | grep "Loaded image" | awk '{ print $3 }')
   elif [ "$containerRuntime" == "podman" ]
   then
-    imageName=$(podman load < "$AGENTDIR"/agent-lcm/idm-agcs-agent-framework.dockerize_agent.tar.gz | grep "Loaded image(s):" | awk '{ print $3 }')
-    if [ "$imageName" == "" ]
-      then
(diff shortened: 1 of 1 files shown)


tanmaygarg-oracle (dismissed):
LGTM!

Djelibeybi:
It would good if you could replace `dockerize` with `containerize` as Docker is a registered trademark and not really being used properly here. However, that's not a blocker either. 

Can you explain why this is necessary? Maybe I'm not spotting it, but it looks like you're trying to set the variable and if if doesn't work, you're doing the same thing again. I'm just not seeing the logic here.

Djelibeybi:
Also, I just noticed line 77 which needs to be fixed ASAP. I'm sorry for not catching it in previous reviews, but that `chmod` value is unacceptable. It can enable group write at _most_ but you should never make directories world writeable and executable.

Djelibeybi:
Finally, a reminder that both Docker and Podman are product names and should be capitalised when used as such, i.e. in any customer visible output or documentation that isn't demonstrating the use of the command-line tools. Please review the content of your error messages to ensure that you're referring to each product correctly.

rishiagarwal-oracle:
> It would good if you could replace `dockerize` with `containerize` in the filename of your tarball as Docker is a registered trademark and not really being used properly here. However, that's not a blocker either.

dockerize comes in the name of the file generated by the OCI build service. We didn't try to change it. Changing of file name would have impact for existing customers and hence we can take it up separately later

> 
> Can you explain why this is necessary? Maybe I'm not spotting it, but it looks like you're trying to set the variable and if if doesn't work, you're doing the same thing again. I'm just not seeing the logic here.

The change here is **Loaded image(s):** vs **Loaded image:** in the podman output

rishiagarwal-oracle:
> Finally, a reminder that both Docker and Podman are product names and should be capitalised when used as such, i.e. in any customer visible output or documentation that isn't demonstrating the use of the command-line tools. Please review the content of your error messages to ensure that you're referring to each product correctly.

Fixed in the output message

rishiagarwal-oracle:
> Also, I just noticed line 89 which needs to be fixed ASAP. I'm sorry for not catching it in previous reviews, but that `chmod` value is unacceptable. It can enable group write at _most_ but you should never make directories world writeable and executable.

@Djelibeybi we will take this up separately. A customer is waiting for the current fix and it needs to go

Djelibeybi:
> The change here is **Loaded image(s):** vs **Loaded image:** in the podman output

Why not just `grep` for `Loaded image` then, which would match either `Loaded image:` or `Loaded image(s):`? You're using the same `awk` for each, so that should work.

Djelibeybi:
Note that this is a change to a sample script, so you can provide the customer with a fixed version before this is merged. Alternatively, the customer could make the modification themselves locally too. On that note, please open a fix for the identified security issue ASAP if you're not going to address it in this PR.

Djelibeybi:
The merge is also blocked due to linter issues, which you'll need to address.

rishiagarwal-oracle:
> 

resolved

tanmaygarg-oracle:
filed issue https://github.com/oracle/docker-images/issues/2510 to track linter issue

Djelibeybi (approved):

rishiagarwal-oracle:
> The merge is also blocked due to linter issues, which you'll need to address.

Are we supposed to fix issues for all severity levels including : style, info and warnings?  We see nothing was reported as an **error** in the scan. What is the gating criteria here?

Djelibeybi:
The gating is warnings and errors (yellow and red). The info/style stuff doesn't actually block a merge, though should be reviewed and deemed acceptable, you should add a directive so that the linter ignores it in future.

Djelibeybi:
> The info/style stuff doesn't actually block a merge

Following up because this is not correct: the Super Linter check will block merging for all `shellcheck` issues.

PR #2523, opened 2022-12-07
Oracle GoldenGate in a Container

sbalousek:
- Allow GoldenGate images to be created with an alternate base image
 - Minor corrections to documentation

### DIFF
--- a/OracleGoldenGate/21c/README.md
+++ b/OracleGoldenGate/21c/README.md
@@ -37,7 +37,7 @@ For more information about Oracle GoldenGate please see the [Oracle GoldenGate 2
 
 ## Build an Oracle GoldenGate Container Image
 
-Once you have downloaded the Oracle GoldenGate software, a container image can be created using the Docker command-line interface.
+Once you have downloaded the Oracle GoldenGate software, a container image can be created using container management command-line applications.
 A single `--build-arg` is needed to indicate the GoldenGate installer which was downloaded.
 
 To create a container image for GoldenGate for Oracle Database, use the following script:
@@ -53,9 +53,7 @@ Similarly, for other Databases like BigData, MySQL, PostgreSQL, etc. provide the
 
 ### Changing the Base Image
 
-By default, the base image used by Docker to build Oracle GoldenGate Docker image is `oraclelinux:8`, but this can be changed.
-
-Changing the base image for the new Docker image is done with the `BUILD_IMAGE` Docker build argument, like in this example:
+By default, the base container image used to create the Oracle GoldenGate container image is `oraclelinux:8`. This can be changed using the `BUILD_IMAGE` build argument. For example:
 
 ```sh
 docker build --tag=oracle/goldengate:21.3.0.0.0 \
@@ -63,7 +61,10 @@ docker build --tag=oracle/goldengate:21.3.0.0.0 \
              --build-arg INSTALLER=213000_fbo_ggs_Linux_x64_Oracle_services_shiphome.zip .
 ```
 
+Oracle GoldenGate 21c requires a base container image with Oracle Linux 8 or later.
+
 ## Running Oracle GoldenGate in a Container
+
 Use the `docker run` command to create and start a container from the Oracle GoldenGate container image.
 
 ```sh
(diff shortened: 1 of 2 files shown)


Djelibeybi (approved):

Djelibeybi:
Could you please add an ignore directive so that the linter check passes?

sbalousek:
The linter check is passing now. Is there Anything else you would like to see here, @Djelibeybi ?

Djelibeybi:
Nope, but until you push the "Re-request review" button (which I just did), I don't get notified that it has been updated. :) Reviewing now.

OracleGoldenGate/21c/README.md
```diff
@@ -41,18 +44,30 @@ To create a container image for GoldenGate for Oracle Database, use the followin
...
 ...
 Successfully tagged oracle/goldengate:21.3.0.0.0
 ```
 Similarly, for other Databases like BigData, MySQL, PostgreSQL, etc. provide the name of the zip file for the INSTALLER argument.
 
+### Changing the Base Image
+
+By default, the base image used by Docker to build Oracle GoldenGate Docker image is `oraclelinux:8`, but this can be changed.
+
+Changing the base image for the new Docker image is done with the `BUILD_IMAGE` Docker build argument, like in this example:
```
Djelibeybi, line 58: Please remove "Docker" from this sentence, e.g. "Changing the base image for the new container image is done using the `BUILD_IMAGE` argument when building the image. For example:"

OracleGoldenGate/21c/README.md
```diff
@@ -41,18 +44,30 @@ To create a container image for GoldenGate for Oracle Database, use the followin
...
+               --build-arg INSTALLER=213000_fbo_ggs_Linux_x64_Oracle_services_shiphome.zip .
 Sending build context to Docker daemon
 ...
 Successfully tagged oracle/goldengate:21.3.0.0.0
 ```
 Similarly, for other Databases like BigData, MySQL, PostgreSQL, etc. provide the name of the zip file for the INSTALLER argument.
 
+### Changing the Base Image
+
+By default, the base image used by Docker to build Oracle GoldenGate Docker image is `oraclelinux:8`, but this can be changed.
```
Djelibeybi, line 56: Please remove "Docker" from this sentence, i.e. "By default, the base image used by the Oracle GoldenGate container image is `oraclelinux:8` but this can be changed."

It would also be good to provide either a list of valid base images or criteria by which someone could determine if something would work as a base image.

OracleGoldenGate/21c/README.md
```diff
@@ -143,13 +158,15 @@ Oracle GoldenGate Administration Client for Oracle
 Version 21.3.0.0.0 ...
 ```
 
-### Running scripts before setup and on startup
-The container images can be configured to run scripts before setup and on startup. Currently, `.sh` extensions are supported. For setup scripts just mount the volume `/u01/ogg/scripts/setup` or extend the image to include scripts in this directory. For startup scripts just mount the volume `/u01/ogg/scripts/startup` or extend the image to include scripts in this directory. Both of those locations are static and the content is controlled by the volume mount.
+### Running Scripts Before Setup and on Startup
+
+The container images can be configured to run scripts before setup and on startup. Currently, `.sh` extensions are supported. For setup scripts just mount the volume `/u01/ogg/scripts/setup` or extend the image to include scripts in this directory. For startup scripts just mount the volume `/u01/ogg/scripts/startup` or extend the image to include scripts in this directory. Both of those locations
```
Djelibeybi, line 163: I would rephrase this as, "Containers created from this image are configured to automatically run scripts found at specific locations at particular points during initial container creation and on subsequent startups."

Then you'd have two blocks: "To run a script before the initial setup of Oracle GoldenGate ......" and  "To run a script each time the container starts but before Oracle GoldenGate is started...."

OracleGoldenGate/21c/README.md
```diff
@@ -143,13 +158,15 @@ Oracle GoldenGate Administration Client for Oracle
...
+are static and the content is controlled by the volume mount.
 
 The example below mounts the local directory myScripts to `/u01/ogg/scripts` which is then searched for custom startup scripts:
```
Djelibeybi, line 166: myScripts needs to be wrapped in back ticks and is missing a leading forward slash.

Djelibeybi (changes requested):

OracleGoldenGate/21c/README.md
sbalousek: I agree the language here is not the best, but the changes in this PR are limited to:

1. Fix capitalization on the section title
2. Shorten lines to make super-linter happy.

Djelibeybi (approved):

Issue #2556, opened 2023-02-03
Checking Docker version(Docker 23.0.0)


Lyn265:
in the BuildContrainerImage.sh ,There is a problem with the value of the docker version of 23.0.0?

error message:Docker version is below the minimum required version 17.09.

remi9898:
I got the same error.  It happens because script BuildContrainerImage.sh removes the dots from the version numbers in order to compare simple integers.  In my case, Docker version 23.0 becomes integer 230, and the minimum allowed version of Docker 17.09 becomes integer 1709.  Well, 230 is always less than 1709 so the test fails even though version 23.0 is much higher than 17.09.  Quick workaround is to edit BuildContrainerImage.sh and change the min allowed version from 17.09 to 18.0 or any other 3-digit number.  Longer term, we need to do better than dot removal comparisons.

alexsuter:
Yep. I face the same issue.

adanecki:
As an alternative, we cloud also just get the full version string, which in this case is 23.0.0 and from it remove dots and cut to 4. 
I don't see the point of cutting the first 5 characters from the version string, as then it's compared to 4 digit integer. No matter what will be the integer, it will always be larger than the minimal version 1709.

DOCKER_VERSION=$("${CONTAINER_RUNTIME}" version --format '{{.Server.Version}}'|| exit 0)
DOCKER_VERSION=$(printf "%.4s" "${DOCKER_VERSION//./}")

brunos1:
> 
I just got this same error and have checked the BuildContrainerImage.sh and just realized it solves the problem.
But instead using "%.4s" I just used "%.6s"

alexsuter:
PR is ready. Can we merge it?

aditya059:
Fixed

PR #2558, opened 2023-02-07
fix: value for formatting the version number

brunos1:
According to issue #2556, the value `"%.5s"` in line 90,for formatting the version number given, doesn't verify all numbers needed and cuts the last value of given string, like number version of `23.0.0`.  Executing the line 
`/usr/bin/docker version --format '{{.Server.Version | printf "%.5s" }}'`  corresponding to the original line 90 in the script, we have: `23.0.`
Which make the condition fails.
Using `"%.8s"` the condition works properly. Any number like 23.45.01 will be verified.


alexsuter:
@brunos1 thx!

aditya059:
Hi @brunos1 

Please sign the Oracle Contributor Agreement (OCA) so that we can merge your pull request.

PR #2570, opened 2023-02-23
Fix version check for docker 23+

felipecrs:
Since version 23 of Docker, they switched away from CalVer to SemVer, which means their versions no longer looks like `23.01` but instead `23.1`.

The buildContainer script fails with docker v23 with:

```console
$ ./buildContainerImage.sh -v 21.3.0 -x -t oracle:xe-21.3.0
WARNING: No swap limit support
Checking Docker version.
Docker version is below the minimum required version 17.09
Please upgrade your Docker installation to proceed.
```

This changes the logic on how the version of docker was being checked so that it works for both the older and the new format.


yunus-qureshi:
@aditya059 please review this PR

Issue #2579, opened 2023-03-01
Feature: Add oraclelinux8-redis dockerfile


jakegt1:
Redis is included in OL8, so is easy to setup by dnf install -y redis.

Djelibeybi:
Resolved via #2580

PR #2580, opened 2023-03-01
#2579 - Add redis dockerfile

jakegt1:


OracleLinuxDevelopers/oraclelinux8/redis/5/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2020, 2022 Oracle and/or its affiliates.
```
AmedeeBulle, line 1: As it is a new file it should be (c) 2023
```suggestion
# Copyright (c) 2023 Oracle and/or its affiliates.
```

.github/workflows/build-and-push-dev-images.yml
```diff
@@ -28,13 +28,13 @@ on:
         required: false
       lang:
         description: List of languages to build
-        default: 'gcc-toolset, golang, nginx, nodejs, php, python, ruby'
+        default: 'gcc-toolset, golang, nginx, nodejs, php, python, ruby, redis'
```
AmedeeBulle, line 31: It's a detail, but we try to respect alphabetical order...
```suggestion
        default: 'gcc-toolset, golang, nginx, nodejs, php, python, redis, ruby'
```

.github/workflows/build-and-push-dev-images.yml
```diff
@@ -28,13 +28,13 @@ on:
...
         required: false
 
 # Default values for the builds triggered by the push event
 env:
   ol: 'oraclelinux7, oraclelinux8'
-  lang: 'gcc-toolset, golang, nodejs, nginx, php, python, ruby'
+  lang: 'gcc-toolset, golang, nodejs, nginx, php, python, ruby, redis'
```
AmedeeBulle, line 37:
```suggestion
  lang: 'gcc-toolset, golang, nodejs, nginx, php, python, redis, ruby'
```

AmedeeBulle (changes requested):
A couple of comments...

- shouldn't we `expose` port 6379?
- can we log on `stdout` instead of `/var/log/redis/redis.log`?

And a question: any good reason to create an `/etc/docker.conf`file rather than editing the `/etc/redis.conf`?

.github/workflows/build-and-push-dev-images.yml
jakegt1: Will do :)

jakegt1:
> A couple of comments...
> 
> * shouldn't we `expose` port 6379?
> * can we log on `stdout` instead of `/var/log/redis/redis.log`?
> 
> And a question: any good reason to create an `/etc/docker.conf`file rather than editing the `/etc/redis.conf`?

Can do these things, all logical. With not editing /etc/redis.conf, thought I would prefer to keep the original for prosperity, but guess it saves a bit of space.

jakegt1:
Have also addressed your comments

OracleLinuxDevelopers/oraclelinux8/redis/5/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2023 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+FROM ghcr.io/oracle/oraclelinux:8
+
+RUN dnf install -y redis && \
+    rm -rf /var/cache/dnf
```
Djelibeybi, line 7:
```suggestion
RUN dnf -y module enable redis:5 && \
    dnf -y install redis && \
    rm -rf /var/cache/dnf
```

This specifically enables the `redis:5` AppStream instead of relying on defaults which have an annoying tendency of sometimes not being what you expect them to be.

OracleLinuxDevelopers/oraclelinux8/redis/5/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN /bin/bash -o pipefail -c "sed 's/protected-mode yes/protected-mode no/g' /etc/redis.conf | \
+    sed 's/^bind/#bind/g' - | sed 's/^logfile .*/logfile \"\"/g' - > /etc/docker.conf"
```
Djelibeybi, line 10:
```suggestion
SHELL ["/bin/bash", "-c", "-o", "pipefail"]

RUN sed -i.bak \
    -e 's/protected-mode yes/protected-mode no/g' \
    -e 's/^bind/# bind/g' \
    -e 's/^logfile .*/logfile \"\"/g' \
    /etc/redis.conf
```

"Docker" is a trademark of Docker, Inc and thus can only be used when referring specifically to Docker, Inc. or one of their products, so the suggestion is to make the changes in-place to `/etc/redis.conf` instead while creating a backup  of the original config too. 

I moved the bash stuff to a `SHELL` directive to make things easier to read and should go before all the `RUN` directives so that it applies to all of them.

btw, I did raise an eyebrow at the combination of disabling protected mode and binding to all interfaces, but then saw the official image does it too (by patching the source code, no less) so I guess it's fine.

Djelibeybi (changes requested):
Just a couple of tiny things, but otherwise good. Are you planning on submitting one for Redis 6 too?

OracleLinuxDevelopers/oraclelinux8/redis/5/Dockerfile
AmedeeBulle: Note that we don't need the `pipefail` stuff anymore since the pipe is gone ;)

Djelibeybi: Ha, I totally missed that. 😂 That's what I get for reviewing PRs before 8am in the morning.

jakegt1: I didn't actually know the SHELL directive existed.. will keep in mind for future.

Also didn't realise there was a dnf module for Redis, so will change that too.

jakegt1: Have made the sed changes and added a 6/Dockerfile for both versions. Nice

AmedeeBulle (approved):
LGTM

Tested both on aarch64 (basic `redis-cli` testing)

Djelibeybi (approved):

Djelibeybi:
Mergified. Nice work, team. 👍🏻 😀

PR #2592, opened 2023-03-16
Resolved docker version 23.0 issue

aditya059:
Previous Line 90 of buildContainerImage.sh will work for all docker versions before 23.0. Min version of docker as per code is 17.09, which is converted to 1709 by the code before comparison. For versions before 23.0, taking 5 letters works. But from docker version 23.0 onwards, 23.0.0, 23.0.1, etc. are converted to 23.0. (when 1st 5 letters are taken) and it becomes 230 after removing the decimal point as per code logic. So, the comparison of 230 > 1709 fails. Removing printf "%.5s" will resolve this.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/buildContainerImage.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/buildContainerImage.sh
@@ -87,7 +87,7 @@ checkPodmanVersion() {
 checkDockerVersion() {
   # Get Docker Server version
   echo "Checking Docker version."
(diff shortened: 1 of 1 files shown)


yunus-qureshi (approved):

yunus-qureshi:
Looks like the lint check here is a false negative

gvenzl:
Great, thanks a lot @yunus-qureshi and @aditya059.

@yunus-qureshi, do you have the list of issues that have been fixed with this PR or do we need to skim through them?

yunus-qureshi:
Thanks @gvenzl. We'll skim and close the issues and redunant PRs.

gvenzl:
@yunus-qureshi, great, thank you!

PR #2628, opened 2023-05-01
Limit the linter to only block on shellcheck errors and warnings

Djelibeybi:


Djelibeybi:
@gvenzl this should reduce the amount of linter issuses

gvenzl (approved):

PR #2634, opened 2023-05-10
add haproxy dockerfile

bgraef:
oracle#2633


OracleLinuxDevelopers/oraclelinux8/haproxy/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2023 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+FROM ghcr.io/oracle/oraclelinux:8
+
+RUN \
+    dnf -y install --setopt=tsflags=nodocs haproxy && \
```
Djelibeybi, line 7: We can't use the `nodocs` option because that suppresses the installation of the `LICENSE` file and we have a legal obligation to provide this file.

OracleLinuxDevelopers/oraclelinux9/haproxy/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2023 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+FROM ghcr.io/oracle/oraclelinux:9
+
+RUN \
+    dnf -y install --setopt=tsflags=nodocs haproxy && \
```
Djelibeybi, line 7: Same as OL8: please remove the `setopt` here.

Djelibeybi (changes requested):
I would suggest adding `EXPOSE` directives for the default ports used by HAProxy so that ports are opened automatically even if not specified when creating or starting the container.

Djelibeybi:
Can you also test/confirm that logging is visible when running `docker logs`, i.e. that output for both `stdout` and `stderr` is sent to the console/foreground?

OracleLinuxDevelopers/oraclelinux8/haproxy/Dockerfile
bgraef: added the `EXPOSE` directives and removed the `nodocs` option.

OracleLinuxDevelopers/oraclelinux9/haproxy/Dockerfile
bgraef: did the same for OL9

bgraef:
Yes, the logging is visible when running `docker logs`

Djelibeybi:
I've just noticed that the file layout is not valid: you need to add a version directory, i.e. `oraclelinux8/haproxy/1.8/Dockerfile` so that the automation can create the right tags. You can create a symlink named `latest` to the latest available version to create that tag too.

Djelibeybi:
You can enable Actions on your fork of this repo to test that the action at least picks up and builds the container image. It will push the result to your own profile, not here.

bgraef:
version 1.8 added for OL8 and version 2.4 added for OL9 to match the package installed

Djelibeybi (approved):

PR #2664, opened 2023-07-12
Tcps custom certs feature

vatsrahul:
Feature to give support to use custom certs in connecting to SIDB via tcps.
Till now by default self-signed verts were used.

With this, If certs directory is mounted in container then script detect that and tcps connection is established using those certs.


gvenzl (approved):
These changes look good to me.
Will merge once @yunus-qureshi also sanctioned it.

yunus-qureshi (approved):

PR #2665, opened 2023-07-13
Update hostname in runOracle.sh and configTcps.sh

viragtripathi:
Resolution for hostname command not found.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/19.3.0/configTcps.sh
+++ b/OracleDatabase/SingleInstance/dockerfiles/19.3.0/configTcps.sh
@@ -31,12 +31,12 @@ ${WALLET_PWD}
 EOF
 
     # Add the certificate
-    orapki wallet add -wallet "${CLIENT_WALLET_LOC}" -trusted_cert -cert /tmp/"$(hostname)"-certificate.crt <<EOF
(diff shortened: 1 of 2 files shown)


yunus-qureshi:
@viragtripathi please clarify if this for the ARM64 image?

viragtripathi:
> @viragtripathi please clarify if this for the ARM64 image?

@yunus-qureshi yes this is for the ARM64 image and I think we should also check for arch type then apply the hostname fix for arm64.

yunus-qureshi:
> > @viragtripathi please clarify if this for the ARM64 image?
> 
> @yunus-qureshi yes this is for the ARM64 image and I think we should also check for arch type then apply the hostname fix for arm64.

Thanks @viragtripathi we are going fix this differently. @aditya059 will raise a PR. Thanks

viragtripathi:
> > > @viragtripathi please clarify if this for the ARM64 image?
> > 
> > 
> > @yunus-qureshi yes this is for the ARM64 image and I think we should also check for arch type then apply the hostname fix for arm64.
> 
> Thanks @viragtripathi we are going to fix this differently. @aditya059 will raise a PR. Thanks

Sounds good @yunus-qureshi. Thanks for the quick response.

aditya059:
This issue will be fixed as part of PR #2666

PR #2670, opened 2023-07-18
Update JDK 11 Dockerfile Checksums

harrisd19:
Updating checksums for version 11.0.20 JDK downloads as per oracle website.

https://www.oracle.com/a/tech/docs/11-0-20-checksum.html


aureliogrb:
Thanks for the pull request @harrisd19.  It was correct, but was superseded by https://github.com/oracle/docker-images/pull/2671/ which include this and other updates before this one was reviewed.

PR #2678, opened 2023-08-02
Fixing bashrc creation in incorrect loc 21.3xe

aditya059:
Resolved bashrc creation in wrong location issue in 21.3xe database. It is supposed to be created inside /home/oracle/ and not inside /root.

### DIFF
--- a/OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile.xe
+++ b/OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile.xe
@@ -79,7 +79,7 @@ RUN cd $INSTALL_DIR && \
     rm -rf $INSTALL_DIR && \
     $ORACLE_BASE/oraInventory/orainstRoot.sh && \
     $ORACLE_HOME/root.sh && \
-	echo 'export ORACLE_SID=XE' >> /home/oracle/.bashrc && \
+    echo 'export ORACLE_SID=XE' >> /home/oracle/.bashrc && \
     chown oracle.oinstall /home/oracle/.bashrc
 
 USER oracle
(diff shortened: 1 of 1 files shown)


yunus-qureshi (dismissed):

yunus-qureshi (approved):

OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile.xe
```diff
@@ -79,12 +81,15 @@ RUN cd $INSTALL_DIR && \
...
+    chown oracle.oinstall /home/oracle/.bashrc
 
 USER oracle
 WORKDIR /home/oracle
 
 HEALTHCHECK --interval=1m --start-period=5m --timeout=30s \
    CMD "$ORACLE_BASE/$CHECK_DB_FILE" >/dev/null || exit 1
 
-CMD exec $ORACLE_BASE/$RUN_FILE
+EXPOSE 1521/tcp 5500/tcp
```
gvenzl, line 93: Why are we adding `EXPOSE` back in? We removed this years ago for good reasons.

gvenzl (changes requested):

OracleDatabase/SingleInstance/dockerfiles/21.3.0/Dockerfile.xe
aditya059: Hi @gvenzl 
Removed it. Added for easy of use / convenience. Didn't knew that it was removed earlier. Do you know the reason why it was removed? If yes, can you give the reason in brief.

gvenzl: `EXPOSE` and `VOLUME` cannot be undone in downstream images. So if anybody uses this image as a base image, the ports stay exposed (or volume if `VOLUME` was used), which may not always be desired. On the flip side, neither `EXPOSE` nor `VOLUME` are needed to expose any port or location inside the container. Users can expose any ports via the `-p` option.

gvenzl (approved):

yunus-qureshi:
Approved, thanks

felipecrs:
I'm trying to understand: why is there a need for a bashrc at all? Isn't it a lot simpler to just `ENV ORACLE_SID=XE` in the Dockerfile?

Also, the removal of `exec` in the `CMD` is **definitely** a problem for signal handling.

Issue #1990, opened 2021-06-23
Reduce image size for a patched OracleDB


beegee1:
Hello everyone,

the image size of a patched Oracle DB is about twice that of an unpatched OracleDB. This is because nearly each file of the original DB installation will be replaced by installing a release update (for example 19.03 to 19.11).
Original 19.03 installation: around 7 GB image size
Patched 19.11 installation: around 15 GB image size

I'm using this image extension: https://github.com/oracle/docker-images/tree/main/OracleDatabase/SingleInstance/extensions/patching

Could the image size be reduced for example by using the oraclelinux base image and copy everything of the patched Oracle installation to the new image? Moreover, after patching these files could be removed: https://github.com/oracle/docker-images/blob/main/OracleDatabase/SingleInstance/dockerfiles/19.3.0/installDBBinaries.sh#L59

Furthermore for installing 19.11 an update of OPatch was necessary, could the update/replacement of OPatch be included in the Dockerfile script? OPatch zip must be provided on your own of course.

abhisbyk:
@beegee1 We have a provision to update OPatch. Download the OPatch zip i.e. Patch 6880880 and place it to `/extensions/patching/patches/one_offs` directory along with other patches if any. Then after extending the image, Opatch will be updated to latest version and all the remaining patches will be applied.

 Additionally, I will be working on reducing the size of patched image by using oracle-linux base image. Will let you know once I reach a conclusion. Thanks.

thelateperseus:
@abhisbyk is there any update on reducing the patched image size? I need to apply patch sets 34133642 and 34086870. These need to be applied as separate `docker build` executions, so the resulting image size is an astronomical 22GB.

abhisbyk:
@thelateperseus You can build these patch sets in one execution itself. You should put 34133642 in `/extensions/patching/patches/release_update` folder and 34086870 in `/extensions/patching/patches/one_offs` folder, and build these patches in one go. It should reduce the size of the image.

Please let me know how it goes.

thelateperseus:
@abhisbyk sorry for the late reply, I've just gotten back to working on this. Thanks for the tip about applying the patches in a single execution.

I've had a go at reducing the size of the patched images. Rather than applying the patches as a separate `docker build` step, I've altered the main `Dockerfile` to apply the patches in the same layer as the DB binaries are installed. The resulting patched image size for 19.16 is 7GB, which is the same as the original 19.03 image. See below for the changes I made to achieve this.

[...]

PR #2682, opened 2023-08-17
Reducing image size for patched db

aditya059:
This PR is to fix issue #1990 (Reduce image size for a patched OracleDB).


yunus-qureshi (approved):
Looks great

gvenzl (approved):

PR #2683, opened 2023-08-17
Tcps cert order

vatsrahul:
Changed order of certs needed from user in cert.crt file.
Now user will follow standard way of having client, intermediates and root certificate in respective order.

Tested with 19.3, 21.3 and 23.2


yunus-qureshi (approved):
Looks good

gvenzl (approved):

PR #2684, opened 2023-08-18
removed -a flag to get container in running state

tanmaygarg-oracle:
removed -a flag to get the containers that are in running state. -a flag lists stopped containers as well. So, exec fails for stopped continers.


rishiagarwal-oracle (approved):
LGTM!

PR #2729, opened 2023-11-08
idm_23.4.1_release

manjunathdhegde-2910:


OracleIdentityGovernance/dockerfiles/12.2.1.4.0/container-scripts/set_feurl.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/python
```
Djelibeybi, line 1: Oracle Linux doesn't support unversioned Python, so this is unlikely to work. If you don't invoke this scrpt as an executable and only call it via the `python3` interpreter, you can/should remove this line. If you do want to call it like an executable, it needs to refer to `python3`

manjunathdhegde-2910: Have fixed both linter issue comment and python comment.

Djelibeybi (approved):

Djelibeybi:
@rishiagarwal-oracle you still need to reapprove this PR since changes have been made after your previous approval.

rishiagarwal-oracle (approved):

PR #2749, opened 2024-01-21
compose file changes

saurabhaahujaa:


OracleDatabase/RAC/OracleRealApplicationClusters/dockerfiles/19.3.0/setupGridEnv.sh
```diff
@@ -192,22 +192,43 @@ fi
...
+        # shellcheck disable=SC2086
+        cp $SCRIPT_DIR/$RESET_FAILED_UNITS  /var/tmp/$RESET_FAILED_UNITS
+        chmod 755 /var/tmp/$RESET_FAILED_UNITS
+        print_message "Setting Crontab"
+        # shellcheck disable=SC2016
+        cmd='su - $GRID_USER -c "sudo crontab $SCRIPT_DIR/$SET_CRONTAB"'
+        # shellcheck disable=SC2086
+        eval $cmd
+        # shellcheck disable=SC2181
+        if [ $?  -eq 0 ];then
```
Djelibeybi, line 225: Instead of disabling three different checks, you could combine these lines into a single line that doesn't require disabling any of themm, i.e. 

```shell
if su - "$GRID_USER" -c "sudo crontab $SCRIPT_DIR/$SET_CRONTAB"; then
   print_message "Successfully installed $SET_CRONTAB using crontab"
else
   error_exit "Error occurred in crontab setup"
fi
```

OracleDatabase/RAC/OracleRealApplicationClusters/dockerfiles/19.3.0/setupGridEnv.sh
```diff
@@ -192,22 +192,43 @@ fi
...
+                                print_message "$SERVICE_NAME stopped."
+                        else
+                                print_message "$SERVICE_NAME is responsive. No action needed."
+                        fi
+                else
+                        print_message "$SERVICE_NAME is not running."
+                fi
+
+        # shellcheck disable=SC2086
+        cp $SCRIPT_DIR/$RESET_FAILED_UNITS  /var/tmp/$RESET_FAILED_UNITS
```
Djelibeybi, line 217: Is there a particular reason you don't just add quotes instead of disabling the check, i.e.

```shell
cp "$SCRIPT_DIR/$RESET_FAILED_UNITS"  "/var/tmp/$RESET_FAILED_UNITS"
```

Djelibeybi (changes requested):
I'm also curious about the `healthcheck` for the various containers in the `docker-compose.yml` files. To me, the command would only determine if process at some point output a particular log entry, it wouldn't be a valid test to run over and over to determine if that service was _still_ running, which is what a health check is meant to do.

saurabhaahujaa:
> I'm also curious about the `healthcheck` for the various containers in the `docker-compose.yml` files. To me, the command would only determine if process at some point output a particular log entry, it wouldn't be a valid test to run over and over to determine if that service was _still_ running, which is what a health check is meant to do.

HealthChecks are meant to set dependency between DNS, RAC and CMAN container. If healthcheck if DNS is not passed, RAC  will fail. If healthcheck of RAC is not passed, then bringing up CMAN container will fail. To avoid failures and people reporting incorrect issues, I have added healthcheck.

OracleDatabase/RAC/OracleRealApplicationClusters/dockerfiles/19.3.0/setupGridEnv.sh
saurabhaahujaa: Well, I am OK to quote it, but changing it in last moment for me means I have to redo all tests of what i tested in the weekend and this is urgent customer escalation and we need this to be merged asap. These checks are only detected during upload to github. At the moment, I am not aware of how to selectively run "shellcheck" of changed/delta code only which github is running. If you insist, i can quote it and run one simple test. FYI, we do have improvements plans to merge this week. By today is priority for merge.

Djelibeybi: https://www.shellcheck.net/ can be installed on Windows, macOS and Linux and is available as a plugin for Visual Studio Code as well as all the JetBrains IDEs, as are all the enabled linters which are listed in the contribution guide.

There is also the `lint` script in the root of this repo that will run the checks for you locally. 

Note that this repo is also not a production service and there is no SLA for merging updates. The folks who review and maintain it are doing so in addition to their actual jobs, not as part of it.

Djelibeybi:
> HealthChecks are meant to set dependency between DNS, RAC and CMAN container.

That's not what a health check is for. It's meant to determine if the service running inside a container is actually running or not. You need to change this so that it's a valid check every time it's run, otherwise it's just going to cause unnecessary load because it's going to run every 30 seconds.

saurabhaahujaa:
I understand that healtcheck is to check if service is healthy or not. That's what we are checking  while setting dependency. See code belo. Otherwise, if customer runs docker-compose up -d, compose will bring up all containers including cman without waiting for racdns and racnode1 to be in healthy condition and whole of setup will fail. I am not able to find any other mechanism  in docker compose to set healthy dependency this way. This is discussed & approved by our team for use in Dev environment. If any issue comes, will handle it, customer can always open issues in github repo and we can review later.  I know you are working on extra support, but is it possible to pass this merge rights to someone else, I dont want to put this additional on you when such urgent customer requirements come. FYI @psaini79 

depends_on:
      racdns:
        condition: service_healthy
      racnode1:
        condition: service_healthy

psaini79 (approved):

psaini79:
@saurabhaahujaa Let us discuss and we need to improve health check based on running services.

@Djelibeybi  thanks for your feedback.

Djelibeybi (approved):

PR #2756, opened 2024-02-05
Oracle GoldenGate 23c

sbalousek:
- In version 23c, the Service Manager is a first-class service. Instead of modifying internal data files, use the REST API to adjust the network settings for Service Manager.


sbalousek:
Hi @Djelibeybi ,

We are gearing up for support for OGG 23c in the coming months. It's been a while, so if there is someone else who should be reviewing pull requests, please let me know!

- Steve

Djelibeybi:
> We are gearing up for support for OGG 23c in the coming months. It's been a while, so if there is someone else who should be reviewing pull requests, please let me know!

If you add two or more people to the `CODEOWNERS` file for your content, then they become the reviewers. I'm only required for content that hasn't been claimed using that method.

Djelibeybi (approved):

PR #2829, opened 2024-07-11
Added OracleLinux9 with Python3.12

europ:


OracleLinuxDevelopers/oraclelinux9/python/3.12/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2023 Oracle and/or its affiliates.
```
Djelibeybi, line 1: Please change this to 2024.

Djelibeybi (changes requested):

OracleLinuxDevelopers/oraclelinux9/python/3.12/Dockerfile
europ: fixed

Djelibeybi (approved):

PR #2830, opened 2024-07-15
Add OracleLinux8 with python3.12

jakegt1:
I see that OL9 w/ python 3.12 exists but good to have it for OL8 too. Hope this is ok!


Djelibeybi (approved):

Djelibeybi:
Thanks!

PR #2838, opened 2024-07-25
Allow word splitting proxy variable in Tuxedo build script

### LINKED ISSUE #2837: Oracle Tuxedo build script fails when `https_proxy` is set
## Problem description

The Oracle Tuxedo container image build script `buildContainerImage.sh` does not function as expected when a valid `https_proxy` environment variable is set.

For example:

```shell
$ ./buildContainerImage.sh -v 22.1.0.0.0 -i tuxedo221000_64_Linux_01_x86.zip -m 72945aafe3514cb1edb4b33cc322a97f
```

Results in a failed container image build:

```shell
...
=====================
Proxy settings were found and will be used during build, https_proxy = <redacted>
=====================
"docker build" requires exactly 1 argument.
See 'docker build --help'.

Usage:  docker build [OPTIONS] PATH | URL | -

Build an image from a Dockerfile
Build of Tuxedo Container image failed.
```

The `docker` command usage banner is displayed in the output which would indicate a malformed command was executed.

## Steps to reproduce

1. Set a HTTPS proxy with `export https_proxy=<protocol://host:port>` substituting a valid proxy host URL
2. Run `./buildContainerImage.sh -v 22.1.0.0.0 -i <installer-name> -m <md5-value>`, substituting valid installer name and MD5 digest values

## Expected behaviour

Running the build script should result in the creation of a new container image.

marcransome:
Fixes #2837.

## Changes

This changes allows the `PROXY_SETTINGS` variable to be word split (as is done in [other build scripts](https://github.com/oracle/docker-images/blob/fa24c6d56a6df05a54dd5c75e6f17ce025f2bde6/OracleWebLogic/dockerfiles/buildDockerImage.sh#L179)) rather than treated as a single argument to the `docker` command. This ensures that the `docker` command is constructed correctly to enable the use of the `https_proxy` environment variable.

### DIFF
--- a/OracleTuxedo/core/dockerfiles/buildContainerImage.sh
+++ b/OracleTuxedo/core/dockerfiles/buildContainerImage.sh
@@ -116,7 +116,7 @@ fi
 
 echo "====================="
 
-docker build  ${PROXY_SETTINGS:+"$PROXY_SETTINGS"} -t oracle/tuxedo:latest -t oracle/tuxedo:"${VERSION}" "${VERSION}"/
+docker build  $PROXY_SETTINGS -t oracle/tuxedo:latest -t oracle/tuxedo:"${VERSION}" "${VERSION}"/
 ret=$?
 if [ "$ret" = "0" ]
     then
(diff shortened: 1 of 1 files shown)


Djelibeybi (approved):

Djelibeybi:
Thanks!

PR #2860, opened 2024-09-10
Add OCNE based k8s tool images that supply helm and kubectl.

jakegt1:
Might need some work. Have been making similar things myself, so thought it would be sensible to supply these. Not sure if it would be better to set ENTRYPOINT as "kubectl", but then makes it harder to use helm. Pain.


Djelibeybi:
What's the reason for using `18` in the path? It doesn't seem to match OCNE release version or K8S version or anything immediately obvious...

Also, would it perhaps make more sense to create two images, one for `kubectl` and the other for `helm` thus making each easier to use and both more applicable to both OCNE and possibly OKE?

jakegt1:
Ahh. ol8_olcne18 . But this is 1.8, so i will change that.

Yep, i could separate them. Should i just name them "kubectl" and "helm" as images then? I could use the kube/helm versions if you think that's better.

I have a personal use case where i would want both in the same image, but that's here or there. They could be separate (and no reason they shouldn't be.)

jakegt1:
> It would probably be cool to have a single image, if it were based on the OCI CLI image or incorporated the OCI CLI and SDK so that connectivity to OKE was also possible as well as OCNE.

This is actually the purpose of my current local image. So should i add another one? My only worry is that these files are getting quite duplicated.

Djelibeybi:
> This is actually the purpose of my current local image. So should i add another one? My only worry is that these files are getting quite duplicated.

I would use https://github.com/oracle/docker-images/pkgs/container/oci-cli as the base image, so that would take care of the OCI CLI and SDK. Then just install Helm and Kubectl. 

We'll need to create an OL9 version of that image, which I can do or you can if you're feeling enthusiastic. :)

jakegt1:
Should i include this in the OracleCloudInfrastructure section, or include it in OracleLinuxDevelopers? 

(OL9 vs OL8 - think i'll let you do that just because not too sure how to architecture the change. Unsure how oci-cli is currently tagged in the actions.)

Djelibeybi:
Let's keep it in Oracle Linux Developers. I'll do the OL9 version.

jakegt1:
Have added. Wasn't sure of naming convention here, as really it's ocne.

jakegt1:
Any suggestion with the OCI cli version to use?

Djelibeybi:
> Any suggestion with the OCI cli version to use?

There's a new version released every week, so it's more about how often we rebuild rather than which version to use. 

Currently there is an automated job running elsewhere that triggers rebuilds when the underlying operating system image is updated. I need to think about how best to trigger rebuilds of these images.

jakegt1:
OK - problem is the lint fails with latest (unsurprisingly), so not sure what best to do here?

Djelibeybi:
> OK - problem is the lint fails with latest (unsurprisingly), so not sure what best to do here?

I suggerst add an ignore for that rule to the Dockerfile: https://github.com/hadolint/hadolint?tab=readme-ov-file#inline-ignores

jakegt1:
@Djelibeybi should be good now.

Djelibeybi:
Awesome. I added this to my list of things to do on Monday after I ran out of time this week to review. Sorry about the delay!

OracleLinuxDevelopers/oraclelinux8/ocne-tools/1.8/help.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
Djelibeybi, line 2: Can you add a copyright header to this script please? Or perhaps just write this info out during build instead of running the script on every instantiation of a container.

Djelibeybi (changes requested):

OracleLinuxDevelopers/oraclelinux8/ocne-tools/1.8/help.sh
jakegt1: Done

PR #2867, opened 2024-09-17
Remove version string format on checkDockerversion

DylanBray:
The printf formatting strips away the last digit of the version number resulting in a 3 digit number instead of a 4 digit number.

This causes the version checker to fail as 1709 is not less than 240 eg. minimum version: 17.0.9 vs my install version: 24.0.5.


yunus-qureshi:
Closing the PR. Please sign the OCA and reopen if needed

PR #2880, opened 2024-10-15
updated release versions to the Oct 2024 CPU.  JDK 17.0.13 has a new license so it has to be download before building.

aureliogrb:
JDK 17 is no longer available under no fee-term-and-conditions so it now requires users to download the binary first, as was the case before for JDK 8 and JDK 11.

Update the checksums for 11 and 8, and the download URL for OpenJDK 23.0.1

### DIFF
--- a/OracleJava/11/Dockerfile
+++ b/OracleJava/11/Dockerfile
@@ -19,7 +19,7 @@
 #      $ docker build -t oracle/jdk:11 .
 #
 # This command is already scripted in build.sh so you can alternatively run
-#		$ bash build.sh
+#      $ bash build.sh
 #
 # The builder image will be used to uncompress the tar.gz file with the Java Runtime.
 
@@ -28,8 +28,8 @@ FROM oraclelinux:7-slim as builder
 LABEL maintainer="Aurelio Garcia-Ribeyro <aurelio.garciaribeyro@oracle.com>"
 
 RUN yum install -y gzip tar
-		
-	
+        
+    
 # Default to UTF-8 file.encoding
 ENV LANG en_US.UTF-8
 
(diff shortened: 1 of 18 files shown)


OracleJava/11/Dockerfile
```diff
@@ -47,10 +47,10 @@ RUN set -eux; \
     if [ "$ARCH" = "x86_64" ]; \
     then \
 		mv "$(ls /tmp/jdk-11*_linux-x64_bin.tar.gz)" /tmp/jdk.tar.gz ; \
-        JAVA_SHA256=f50fdec8a48a9b360d30ecc29af36f63f04f0b70ec829d3bf821e4e361682791 ; \
+        JAVA_SHA256=d22d0fcca761861a1eb2f5f6eb116c933354e8b1f76b3cda189c722cc0177c98 ; \
```
rjeberhard, line 50: I'll just mention this once, but you should change the copyright statement in each file to reference 2024.

OracleJava/17/Dockerfile
```diff
@@ -29,30 +28,31 @@ LABEL maintainer="Aurelio Garcia-Ribeyro <aurelio.garciaribeyro@oracle.com>"
...
 ##
+COPY *.tar.gz /tmp/
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN set -eux; \
 	ARCH="$(uname -m)" && \
-	# Java uses just x64 in the name of the tarball
     if [ "$ARCH" = "x86_64" ]; \
-        then ARCH="x64"; \
+    then \
+		mv "$(ls /tmp/jdk-17*_linux-x64_bin.tar.gz)" /tmp/jdk.tar.gz ; \
```
rjeberhard, line 47: Very minor, but this `mv` is spaced over further than the else.

OracleJava/README.md
```diff
@@ -65,9 +65,9 @@ The script `build.sh` will tag the images it creates with the JDK version, and w
 
 ## Licenses
 
-JDK 23, 21, and 17 are downloaded, as part of the build process, from the [Oracle Website](https://www.oracle.com/javadownload) under the [Oracle No-Fee Terms and Conditions (NFTC)](https://java.com/freeuselicense).
+JDK 23 and 21 are downloaded, as part of the build process, from the [Oracle Website](https://www.oracle.com/javadownload) under the [Oracle No-Fee Terms and Conditions (NFTC)](https://java.com/freeuselicense).
 
-For building JDK 11, JDK 8, and Server JRE 8 you must first download the corresponding Java Runtime from the [Oracle Website](https://www.oracle.com/javadownload) and accept the license indicated on that page.
+For building JDK17, JDK 11, JDK 8, and Server JRE 8 you must first download the corresponding Java Runtime from the [Oracle Website](https://www.oracle.com/javadownload) and accept the license indicated on that page.
```
rjeberhard, line 70: "JDK17" -> "JDK 17"

rjeberhard (dismissed):
Very minor comments but nothing that blocks approval.

OracleJava/11/Dockerfile
aureliogrb: Thanks for the reminder.  Updated

OracleJava/README.md
aureliogrb: Fixed

OracleJava/17/Dockerfile
aureliogrb: Fixed in all files/lines

OracleJava/11/Dockerfile
```diff
@@ -28,8 +28,8 @@ FROM oraclelinux:7-slim as builder
 LABEL maintainer="Aurelio Garcia-Ribeyro <aurelio.garciaribeyro@oracle.com>"
 
 RUN yum install -y gzip tar
-		
-	
+        
+
```
Djelibeybi, line 32: If you're cleaning up anyway, please remove the spaces from these lines. There's an option in VS Code (and most editors/IDEs) that automatically remove spaces from the end of lines when saving, which should be enabled.

OracleJava/11/Dockerfile
```diff
@@ -43,48 +43,48 @@ ENV JAVA_HOME=/usr/java/jdk-11
...
     then \
-		mv "$(ls /tmp/jdk-11*_linux-x64_bin.tar.gz)" /tmp/jdk.tar.gz ; \
-        JAVA_SHA256=f50fdec8a48a9b360d30ecc29af36f63f04f0b70ec829d3bf821e4e361682791 ; \
+    mv "$(ls /tmp/jdk-11*_linux-x64_bin.tar.gz)" /tmp/jdk.tar.gz ; \
+        JAVA_SHA256=d22d0fcca761861a1eb2f5f6eb116c933354e8b1f76b3cda189c722cc0177c98 ; \
     else \
-	    mv "$(ls /tmp/jdk-11*_linux-aarch64_bin.tar.gz)" /tmp/jdk.tar.gz ; \
-    	JAVA_SHA256=ff0c3387f47ed2eff9fdf936952573113824fc29de463bb13a3a804cc13d2ae8 ; \    	
+    mv "$(ls /tmp/jdk-11*_linux-aarch64_bin.tar.gz)" /tmp/jdk.tar.gz ; \
+        JAVA_SHA256=3fc0d93f6363d32723c293ba5a9016e8ab27410351ed804020cfe71e87d3bc0a ; \
```
Djelibeybi, line 53: There should never be anything (including spaces) after the `\` character at the end of a line, as this is the line break marker so if something appears after it, it can break the break, and weirdness can occur.

Djelibeybi (changes requested):
The last cleanup task is to ensure that there are no whitespace characters left at the end of any line and that blank lines have no whitespace at all. There are a couple of settings in VSCode (and other editors/IDEs) to do this automatically on save.

OracleJava/11/Dockerfile
aureliogrb: Removed whitespaces

Djelibeybi (approved):

PR #2882, opened 2024-10-18
Make Oracle Database healthcheck wait for the custom scripts to finish

felipecrs:
**This PR is currently incomplete as it still needs to be applied to other Oracle Database versions.**

**I am opening it now to gather initial feedback on the approach.**

We use Oracle Database for some of our CI testing, both via `docker-compose` and Kubernetes, relying on Oracle’s health check to determine when the database is ready for the application to connect.

Our process involves creating users and tables through an `init_db.sh` script, which runs at `startup` (not at `setup`), since we leverage the `prebuiltdb` extension. Therefore, it's crucial that the application only attempts to connect once the database is fully initialized, and all startup scripts have executed.

The current health check only verifies if the database is running, without considering whether the initialization logic has finished. This can result in race conditions where the application tries to connect before the database is fully started.

This PR aims to resolve that by introducing a marker file, which signals that the database has completed its startup process. Only after this file is present will the health checker attempt to connect to the database.

So, what do you think?


felipecrs:
To be fair, the linter is complaining about (real) issues not related with my code. I can still fix them if you want.

yunus-qureshi:
@felipecrs  can we fix the lint errors and resolve the conflicts please

felipecrs:
@yunus-qureshi, done. Please check again.

yunus-qureshi (approved):

PR #2922, opened 2025-02-19
minor link fixes

s1saurabh:


psaini79 (changes requested):
@Suarabh, why we are still using OL7 for dummy NFS storage. I think we can migrate to OL8 as a base image?

s1saurabh:
> @suarabh, why we are still using OL7 for dummy NFS storage. I think we can migrate to OL8 as a base image?

This was to support older releases based on OL7, let me remove it from here and put under older release section

psaini79 (dismissed):

PR #2938, opened 2025-03-05
add cli version tag to build pipeline

### LINKED ISSUE #2935: Add OCI-CLI Version Tag to Docker Images
Currently, the OCI-CLI Docker images are only tagged with latest or the build timestamp (e.g., [20250226](https://github.com/oracle/docker-images/pkgs/container/oci-cli/362868090?tag=20250226)). This makes it difficult to reference a specific CLI version reliably.

I would like to request an additional tag that corresponds to the CLI version itself. For example, the latest version ([latest](https://github.com/oracle/docker-images/pkgs/container/oci-cli/362868090?tag=latest)) currently refers to `3.51.9`, so an additional tag like `3.51.9` would be helpful.

This would allow users to pull a specific CLI version without relying on latest or tracking timestamps.

Expected Behavior:
	•	Each OCI-CLI Docker image should include a tag that corresponds to the CLI version (e.g., 3.51.9).
	•	This tag should be updated alongside latest with each new release.

Would it be possible to implement this in future builds?

frankforpresident:
Resolves #2935 

@Djelibeybi Could you do the testing for me and update where needed?

### DIFF
--- a/.github/workflows/build-and-push-oci-cli-image.yml
+++ b/.github/workflows/build-and-push-oci-cli-image.yml
@@ -3,15 +3,15 @@ name: Build and Push OCI CLI Image
 on:
   workflow_dispatch:
   schedule:
-    - cron: '30 0 * * 3'  # Tuesday@17h30 Pacific = Wednesday@00h30 UTC
+    - cron: '0 * * * *'  # Runs every hour
   push:
     branches: [ "main" ]
     paths:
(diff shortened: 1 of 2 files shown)


Djelibeybi:
@frankforpresident I can only review after you sign the OCA. Otherwise, I'm not even allowed to look at it.

PR #2959, opened 2025-04-15
Add HTTPD OL9 image

jakegt1:
Wanting some review on this, just thinking about default listen ports, should SSL be enabled (i.e dnf install -y mod_ssl), etc.


Djelibeybi:
My immediate reaction was that HTTPS/TLS support would be preferable, but that would require either the creation of self-signed certificates (so that Apache actually starts) or a startup script that checked for the existence of certificates (provided via bind mount or some other mechanism) before enabling TLS support.

Thoughts?

jakegt1:
Have added something simple, and checked that it works.

jakegt1:
Anything else to do here?

OracleLinuxDevelopers/oraclelinux9/httpd/2.4/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -eu pipefail
+
+SERVER_CERT="$CERTIFICATE_DIR/localhost.crt"
+SERVER_KEY="$KEY_DIR/localhost.key"
```
Djelibeybi, line 6: We should also check for the existance of the standard ACME files, e.g. `privkey.pem`, `cert.pem`, `fullchain.pem` and `ca.pem`.

Or just rename these two to `privkey.pem` and `fullchain.pem` so they can be overloaded via bind mount to a host-wide ACME certificate.

jakegt1: This is just the default for the SSL config - I could do a sed to fix it though. Expecting acme mounts seems logical.

Djelibeybi: Also, some sort of output to `stderr` if we do fallback to self-signed certificates would probably be pretty useful too.

Djelibeybi:
Can you please rebase this PR? I'll merge it then.

jakegt1:
Have also made the acme changes (hope they make sense?)

Didn't put a backup file this time.. mainly because i think httpd would still try pick it up.

jakegt1:
Anything more to do here? I think it's good

Djelibeybi (approved):

PR #2993, opened 2025-07-02
Removed the need to pass the OCI region as an argument to the dockerfiles

aureliogrb:
Java Management Service (JMS) has been updated to no longer require users to specify OCI_REGION when downloading releases.   Simplified the dockerfiles by removing that argument


Djelibeybi (approved):
Yay!

PR #3005, opened 2025-08-08
Add golang 1.24.4/latest images for OL9, add 1.24 golang to OL8

jakegt1:
There was only a '1.18' image for OL9, and it appears to weirdly be not acutally 1.18.  Thought i would add a fixed 1.24.4 image (for anyone who wants it), and latest (which is currently also 1.24.4.


jakegt1:
One minor issue - the 'ol8' module is *not* the latest version of golang available to OL8. Not sure what to do about the 'latest' tag.

jakegt1:
No worries!

PR #3025, opened 2025-09-01
Add OL10 developer images

jakegt1:
Quick ports of everything that still works. OCNE does not appear to be in OL10 yet, so removed those. And the gcc-toolset doesn't seem to use scl anymore, so wasn't sure what to do about that (but surely should have a build-tools container eventually.)

### DIFF
--- a/OracleDatabase/RAC/OracleRealApplicationClusters/containerfiles/scripts/oracommon.py
+++ b/OracleDatabase/RAC/OracleRealApplicationClusters/containerfiles/scripts/oracommon.py
@@ -291,7 +291,7 @@ class OraCommon:
                 env_dict[key] = value
                 self.oenv.update_env_vars(env_dict)
              else:
-                msg='''Variable {0} value is not defined to add in the env variables. Exiting!'''.format(value)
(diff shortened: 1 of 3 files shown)


cjbj:
Don't forget to update https://github.com/oracle/docker-images/blob/main/OracleLinuxDevelopers/README.md

OracleLinuxDevelopers/oraclelinux10/python/3.12-oracledb/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2025 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+
+FROM ghcr.io/oracle/oraclelinux10-python:3.12
+
+RUN dnf -y install oraclelinux-developer-release-el10 && \
+    dnf -y install python3.12-oracledb && \
+    # Optionally install Oracle Instant Client to use python-oracledb Thick mode
```
cjbj, line 8: Delete the references to Instant Client - it hasn't been released for OL10.

jakegt1:
@cjbj all good?

PR #3034, opened 2025-10-22
Add valkey to list of images to build

jakegt1:


AmedeeBulle (changes requested):
We need it in `default` as well (line 31)

